PR #20, opened 2015-01-13
New design, localization, remove b&you connector, various fixes

jsilvestre:
Basically this PR cannot be reviewed from Github, sorry!

My advice is that you read the code that I actually changed:

## Client
* start with client/app/initialize.coffee
* then client/app/router.coffee
* then client/app/views/application.coffee
* then client/app/views/menu.coffee with menu_item.coffee

I also removed unused stuff.

I didn't change client/app/views/konnector.coffee (generic form generation), or just a few lines but nothing that deeply changes the way it works.

## Server
You can check everything that has changed here.
* data are now passed within index.html (with tests for the specific thing we do, cf server/models/konnector.cofffee::appendConfigData() \o/)
* server-side localization
* notifications per connector


jsilvestre:
Server-side localization is missing. Don't merge before I add it.

jsilvestre:
I added a few more fixes + the server-side localization. It's ok for me.

PR #27, opened 2015-01-28
Improved: the import button is now disabled when an import is running. Close #26

### LINKED ISSUE #26: "Import" button should be blocked when an import is running

jsilvestre:


frankrousseau:
`onImportClicked` method lacks of comments.

jsilvestre:
I rebased against the latest upstream modifications, and added the comments.

client/app/views/konnector.coffee
```diff
@@ -115,23 +119,39 @@ module.exports = class KonnectorView extends BaseView
...
-                    @$('.error .message').html t(err.responseText)
-                    @$('.error').show()
+    enableImportButton: ->
+        @$('#import-button').removeClass 'disable'
+        @$('#import-button').attr 'aria-busy', false
+        @$('#import-button').attr 'aria-disabed', false
+
+    onImportClicked: ->
+        # don't restart the import if an import is running
+        unless @model.get('isImporting')
```
m4dz: Why don't use a quick return here to prevent pyramid of doom, like `if @model.get 'isImporting' then return` ?

client/app/views/konnector.coffee
```diff
@@ -115,23 +119,39 @@ module.exports = class KonnectorView extends BaseView
...
+            importInterval = $("##{slug}-autoimport-input").val()
+            importDate = $("##{slug}-import-date").val()
+
+            # save field values and start importing
+            data = {fieldValues, importInterval}
+            @model.save data,
+                success: (model, success) =>
+                error: (model, err) =>
+                    # cozycloud.cc timeout is not considered like an error
+                    unless err.status is 504
```
m4dz: Like previously, prefer a quick return rather a new nested level.

client/app/views/styles/_konnector.styl
```diff
@@ -39,6 +39,10 @@
             margin 30px 0 10px 0
             button
                 outline none
+                &.disable
```
m4dz: is the `disable` class only used here? If yes, then prefer styling `&[aria-busy=true]` for cursor and `&[aria-disabled=true]` for greying. It then do not rely on a class that duplicates native state-support provided by aria.

client/app/views/konnector.coffee
jsilvestre: I don't like to do it, so I avoid it when I can. I find quick return to be less readable.

client/app/views/styles/_konnector.styl
jsilvestre: Sure I can change that, I'm just not used to it.

client/app/views/konnector.coffee
m4dz: Humm… interesting, I always think people find it more readable. @frankrousseau any idea / preference on this pattern? (and I'll stick to the majority)

jsilvestre:
@m4dz can we say it's okay because I have more modifications to push. I don't want to put that under the table, I'll make sure we do something about it.

m4dz:
Yep, right for me. I'll keep it in mind for Cozy-Ui :wink:.

PR #30, opened 2015-01-29
Fixed tests and improved save/import code

jsilvestre:
Save field values for a konnector and import a konnector are two server actions done with the same user action.
I made more clear what the button does ("save and import") and separated the code of the "save" and the "import".

Also, I broke the tests a while ago without noticing it, they are now fixed.


poupotte:
It seems good for me (review + fonctional tests).

PR #31, opened 2015-01-29
Autoimport

poupotte:


jsilvestre:
General remark: there is a class, but there are also variables in the file scope. I suggest we move them to the class scope (`timeouts`, `timeout` and `nextUpdates`). Or we don't do a class, which is perfectly acceptable too. There is no best practice yet on how to create singletons (see https://github.com/cozy/cozy-guidelines/pull/8#issuecomment-71809879 and following comments).

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+        if reset
+            nextUpdates = {}
+        if Object.keys(nextUpdates).length is 0
+            Konnector.all (err, konnectors) =>
+                async.eachSeries konnectors, (konnector, callback) =>
+                    # if both importInterval and lastAutoImport are valid
+                    if konnector.importInterval? \
+                    and konnector.importInterval isnt 'none' \
+                    and konnector.lastAutoImport? \
+                    and fs.existsSync path.resolve("server/konnectors/#{konnector.slug}.coffee")
```
jsilvestre: you can use `konnectorHash[konnector.slug]?` to replace the fs check

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
-
-    handleTimeout: (konnector) ->
-
-        # if date is present in fieldValues
+                , (err) =>
+                    # Initialize every day to avoid to have lots of long timeout.
+                    @prepareNextCheck()
+                    if timeout?
+                        clearTimeout timeout
+                    timeout = setTimeout @start.bind(@), day
```
jsilvestre: I think the poller timeout management could be extracted to a function, to avoid duplication and to make it more semantic/understandable. Also that will shorten this function.

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
 format = "DD/MM/YYYY [at] HH:mm:ss"
 periods = {hour: hour, day: day, week: week, month: month}
 timeouts = {}
+nextUpdates = {}
+timeout = null
 
 class KonnectorPoller
 
-    start: ->
+    start: (reset=false, cb=null) ->
```
jsilvestre: I would use `callback` here (see my other comment on the current "callback" variable)

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
-                        nextUpdate = nextUpdate.add interval, 'ms'
-                        log.debug "#{konnector.slug} | Next update : " +
-                        "#{nextUpdate.format(format)}"
-
-                        @prepareNextCheck konnector, interval
+        if reset
+            nextUpdates = {}
+        if Object.keys(nextUpdates).length is 0
+            Konnector.all (err, konnectors) =>
+                async.eachSeries konnectors, (konnector, callback) =>
```
jsilvestre: I would use `next` insteand of `callback` here. It's just semantic.

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+                    # if both importInterval and lastAutoImport are valid
+                    if konnector.importInterval? \
+                    and konnector.importInterval isnt 'none' \
+                    and konnector.lastAutoImport? \
+                    and fs.existsSync path.resolve("server/konnectors/#{konnector.slug}.coffee")
+                        nextUpdate = @findNextUpdate(konnector)
+                        nextUpdates[konnector.slug] = [nextUpdate, konnector]
                         callback()
```
jsilvestre: I suggest you extract the iterator to a class method like `initializeKonnectorPolling` or something (I don't have a good name)

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
-                        interval -= now.valueOf()
-                        log.debug "#{konnector.slug} didn't miss an importation cycle"
-
-                        nextUpdate = now.clone()
-                        nextUpdate = nextUpdate.add interval, 'ms'
-                        log.debug "#{konnector.slug} | Next update : " +
-                        "#{nextUpdate.format(format)}"
-
-                        @prepareNextCheck konnector, interval
                         callback()
```
jsilvestre: You can do only one "callback()" call at the end of the function

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+                    @prepareNextCheck()
+                    if timeout?
+                        clearTimeout timeout
+                    timeout = setTimeout @start.bind(@), day
+                    cb() if cb?
+        else
+            @prepareNextCheck()
+            if timeout?
+                clearTimeout timeout
+            timeout = setTimeout @start.bind(@), day
```
jsilvestre: It's purely subjective here: I like to write "1 * day" so it's readable like "1 day", but don't change it if you think it doesn't improve anything.

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+            cb() if cb?
+
+    findNextUpdate: (konnector) ->
+        importInterval = periods[konnector.importInterval]
+        now = moment()
+        lastImport = moment(konnector.lastImport)
+        lastAutoImport = moment(konnector.lastAutoImport)
```
jsilvestre: useless parenthesis

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+            # calculate the supposed last Auto-import
+            importTime = lastAutoImport.add importInterval, 'ms'
+            # calculate the time elapsed
+            while importTime.valueOf() < now.valueOf()
+                importTime = importTime.add importInterval, 'ms'
+
+            log.debug "#{konnector.slug} | Next update : " +
+            "#{importTime.format(format)}"
+            return importTime
+            callback()
```
jsilvestre: `callback()`?

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
         startDate = konnector.fieldValues.date if konnector.fieldValues.date?
         # Retrive current Autoimport value in database
         Konnector.find konnector.id, (err, savedKonnector) =>
             savedKonnector.injectEncryptedFields()
-
             currentInterval = savedKonnector.importInterval
 
-            # If the importInterval has changed
+            # If the importInterval has changed or it's a new
             if konnector.importInterval isnt currentInterval
```
jsilvestre: We may change the way handleTimeout works, or change how the save/import (in the controller) is done, because based on https://github.com/cozy-labs/konnectors/blob/master/server/controllers/konnectors.coffee#L39, the code cannot enter in that condition. I feared I broke that...confirmed.

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+            @createTimeout konnector, interval
+
+    prepareNextCheck: () ->
+        for slug in Object.keys(nextUpdates)
+            # If interval is more than 1 day,
+            # timeout isn't started (prepareNextCheck call every day)
+            [nextUpdate, konnector]  = nextUpdates[slug]
+            now = moment()
+            interval = nextUpdate.diff now.clone(), 'ms'
+            if interval < day
```
jsilvestre: I think adding a comment here to explain we only poll day after day would be nice.

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
 
-        log.info "Next check of konnector #{konnector.slug} on " +
-        "#{nextUpdate.format(format)}"
-        # Create the timeout and place it timeouts
+    createTimeout: (konnector, interval) ->
         timeouts[konnector.slug] = setTimeout @checkImport.bind(@, konnector, interval), interval
 
-    checkImport: (reference, interval) ->
-        # retrieves the last version of the konnector
-        Konnector.find reference.id, (err, konnector) =>
```
jsilvestre: You removed that, are you sure when the code executes the import, it has the latest version of the konnectors? Anyway, it'd be better to do it in the "importer" rather than here, so it's good to remove it from here.

server/lib/konnector_poller.coffee
```diff
@@ -18,187 +18,154 @@ month = 30 * day
...
+
+    prepareNextCheck: () ->
+        for slug in Object.keys(nextUpdates)
+            # If interval is more than 1 day,
+            # timeout isn't started (prepareNextCheck call every day)
+            [nextUpdate, konnector]  = nextUpdates[slug]
+            now = moment()
+            interval = nextUpdate.diff now.clone(), 'ms'
+            if interval < day
+                @createTimeout konnector, interval
```
jsilvestre: I would extract 
```coffeescript
            now = moment()
            interval = nextUpdate.diff now.clone(), 'ms'
            if interval < day
                @createTimeout konnector, interval
```
Because it's used twice, and one is not commented. No idea about the name the function could have. Maybe `createTimeout` and rename `createTimeout` to `startTimeout` (or merge `createTimeout` to the new `createTimeout`, I don't know)

server/lib/konnector_poller.coffee
```diff
@@ -17,188 +16,175 @@ week = 7 * day
...
-                if konnector.importInterval? \
-                and konnector.importInterval isnt 'none' \
-                and konnector.lastAutoImport? \
-                and fs.existsSync path.resolve("server/konnectors/#{konnector.slug}.coffee")
 
-                    importInterval = periods[konnector.importInterval]
-                    now = moment()
-                    lastImport = moment(konnector.lastImport)
-                    lastAutoImport = moment(konnector.lastAutoImport)
+    constructor: () ->
```
jsilvestre: useless parenthesis

server/lib/konnector_poller.coffee
```diff
@@ -17,188 +16,175 @@ week = 7 * day
 month = 30 * day
 format = "DD/MM/YYYY [at] HH:mm:ss"
 periods = {hour: hour, day: day, week: week, month: month}
```
jsilvestre: a shortcut is: `periods = {hour, day, week, month}`

m4dz:
Ok for me. Just a little simple question introduced by the schedule imports mechanism: why do we do not rely on a queue task manager like [kue](https://github.com/learnboost/kue) to deal with this jobs at the platform level?

jsilvestre:
That's a good question. I think it would be overkill, regarding the simplicity of what we need. A good thing to add to the platform is probably the possibility for applications to be woken up by the platform in order to a specific job. That would avoid the need for applications like konnectors to be up and running all the time.

m4dz:
> A good thing to add to the platform is probably the possibility for applications to be woken up by the platform

That's what I have in mind : app can register against the platform to be wake up, or to record a job for a later call. Then the platform (using a aueue task system, or an internal process) may call the app at registered date. Maybe we could keep this idea (aka: i'll create an issue for this).

jsilvestre:
+1 for the issue creation

PR #33, opened 2015-02-03
Changed: notification behaviour

jsilvestre:
* each konnector is now responsible for creating its success notification
* the failure notification is handled generically
* there is no "successful import" permanent notification anymore
* if there is an error notification, then a latter import is successful, the error notification is removed (or replaced, if the konnector provided a notification content)


frankrousseau:
I don't understand well. Is there only the Commit and Tweet konnectors which send information about import?

jsilvestre:
As of today, yes. The thing is that it won't break with previous behaviour except for the non permanent and annoying notification "it all went okay" if you had enabled the auto-import. If, as I think, this is the right way to handle this (I want to try with the ISEN konnectors which is more complex), I will add notifications for other konnectors.

Since I want to iterate and try to make it as generic as possible, to avoid putting to much things into the konnectors themselves, I want to work on the ISEN konnectors (as I said, it's more complex) on order to acknowledge my implementation or do changes accordingly.

jsilvestre:
I should not have written the post, I'm just going to do it.

jsilvestre:
Done @frankrousseau and @m4dz

frankrousseau:
Just for information. Konnector was a quick and dirty project when I started it. It probably lacks a lot of comments. Feel free to add more comments when you have the opportuniy.

PR #35, opened 2015-02-04
Fixed: error notifications had the prefix twice

jsilvestre:
The output was "Konnector Free: Konnector Free: the notification content".


m4dz:
Woups, I dunno who review the previous PR :yum:…

PR #40, opened 2015-02-20
Fixed: encrypted fields were lost during automatic imports

jsilvestre:


jsilvestre:
@poupotte I haven't managed to fix the tests. I don't understand why they fail. Can you have a look at it to see if you find something quickly? Otherwise I'll go down another path.

PR #44, opened 2015-03-07
Add forex konnector

jankeromnes:
I'm adding a Konnector to import reference exchange rates for currencies, imported from the European Central Bank [website](https://www.ecb.europa.eu/stats/eurofxref/), because we need them for Kresus.

The Konnector is close to working, and I need some help. @frankrousseau could you please have a look?


jankeromnes:
Note: I did some drive-by fixes in the locales.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
+americano = require 'americano-cozy'
+request = require 'request'
+moment = require 'moment'
+crypto = require 'crypto'
+async = require 'async'
```
jankeromnes: Here, I would like to import the [xml2js](https://www.npmjs.com/package/xml2js) npm module. Is there a process to add it to Cozy?

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+# http://www.global-view.com/forex-trading-tools/forex-history/exchange_csv_report.html?CLOSE_1=ON&start_date=1/1/1900&stop_date=03/06/2015&Submit=Get%20Daily%20Stats
+# http://www.global-view.com/forex-trading-tools/forex-history/exchange_csv_report.html?CLOSE_1=ON&CLOSE_2=ON&CLOSE_4=ON&start_date=1/1/1900&stop_date=03/06/2015&Submit=Get%20Daily%20Stats
+# http://www.federalreserve.gov/datadownload/Output.aspx?rel=H10&series=a6a0113179fdeb9c6fbcdc18575ec09c&lastObs=&from=01/01/2000&to=03/07/2015&filetype=csv&label=omit&layout=seriescolumn
+
+# Models
+
+#Currency = americano.getModel 'Currency',
+#    code: String # ISO 4217 currency code
+#    title: String
+#    symbol: String
```
jankeromnes: Maybe this model is not useful, and apps can deal with the currency codes themselves.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+        ILS: "checkbox"
+        INR: "checkbox"
+        KRW: "checkbox"
+        MXN: "checkbox"
+        MYR: "checkbox"
+        NZD: "checkbox"
+        PHP: "checkbox"
+        SGD: "checkbox"
+        THB: "checkbox"
+        ZAR: "checkbox"
```
jankeromnes: Since this commit breaks the Konnectors app, I wasn't able to see how these fields look like. I'm hoping for checkboxes a bit [like here](http://www.global-view.com/forex-trading-tools/forex-history/index.html).

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+
+    # Fetch rates data from the ECB website and save them as Cozy objects.
+    fetch: (requiredFields, callback) ->
+        log.info 'import started'
+
+        url = dailyUrl
+        today = new Date() # FIXME day-precision, not millisecond-precision.
+
+        # Compute the latest rates we know.
+        # FIXME maybe this doesn't need to go through all the rates?
```
jankeromnes: Is there a way to see the `CurrencyRate` with the latest `date` for a given `currency`?

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+        latestRates = {}
+        CurrencyRate.all, (err, rates) ->
+            return callback err if err
+            for rate in rates
+                date = latestRates[rate.currency].date
+                if !date or date < rate.date
+                    latestRates[rate.currency] = rate
+        
+        # Figure out how far back in time we need to fetch.
+        for currency in requiredFields
```
jankeromnes: I hope `requiredFields` doesn't contain more things than just the fields I added earlier (`EUR`, `USD`, etc).

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+            if !requiredFields[currency] #.checked ?
```
jankeromnes: Not sure if this will be `true/false` or if I need to verify a `.checked` attribute.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+                continue
+            date = latestRates[currency].date
+            if !date or date #older than 90days
```
jankeromnes: Not sure how to deal with day-precision dates (as opposed to javascript millisecond-precision dates) in coffescript / cozy yet.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+                # Jackpot! We have to refetch from the beginning.
+                url = histUrl
+                break
+            if date #older than yesterday
```
jankeromnes: Same as above.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+                url = recentUrl
+
+        options =
+            method: 'GET'
+            url: url
+
+        request options, (err, res, body) ->
+            return callback err if err
+
+            result = parseXMLString body # FIXME use npm module xml2js?
```
jankeromnes: This should be something like `xml2js.parseString(body, function(err, result) { ... }` when `xml2js` is actually imported.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,152 @@
...
+                    rate = quote.$.rate
+                    latest = latestRates[currency]
+                    if latest and latest >= date
+                        continue
+                    currencyrate = new CurrencyRate
+                        date: date
+                        rate: rate
+                        currency: currency
+                        base: "EUR"
+                    currencyrate.save (err) ->
```
jankeromnes: Is it ok to save multiple `CurrencyRate` for the same `date` but with different `currency`?

jankeromnes:
Also, I don't know how to properly test my change (e.g. see cozy install logs when things break), and some files (like konnectors/client/public/javascripts/app.js.map) probably need to be rebuilt.

frankrousseau:
ok I will check that soon.

package.json
```diff
@@ -86,6 +86,9 @@
     "Commit": {
       "description": "Commit infos imported from Github."
     },
+    "CurrencyRate": {
+      "description": "Exchange rates importent from the European Central Bank."
```
frankrousseau: Do you mean imported?

server/konnectors/currencies.coffee
frankrousseau: In the repository root directory, run:

    npm install xml2js --save 

It will add the dependency to your local installation and add a new line to the `package.json` file to include it in every installation.

server/konnectors/currencies.coffee
frankrousseau: If you don't use the model locally, it won't be seen by other apps. CouchDB is schemaless. So this schema makes sense only for the current app.

server/konnectors/currencies.coffee
frankrousseau: I think it will be displayed as a long list.

jankeromnes:
Thanks for the feedback and replies! I'll address them one by one in fixup commits, and later squash them back into one commit for final review.

server/konnectors/currencies.coffee
frankrousseau: Yep, here is a tutorial for achieving sorted request with CouchDB: https://github.com/cozy/cozy-data-system/wiki/Couchdb-help#order-docs-with-a-given-field-and-filter-on-this-field

Tell me if it's not clear.

server/konnectors/currencies.coffee
jankeromnes: If I have to run this in my root directory (I suppose that means ssh into my cozy), will that force everyone to do that in their cozy as well? Isn't there a way to add that package to the official cozy distribution?

I added the `package.json` dependency and fixed the code in the temporary follow-up commit "Fix: Add xml2js as dependency."

server/konnectors/currencies.coffee
frankrousseau: It contains every parameters that the user sends.

server/konnectors/currencies.coffee
jankeromnes: Maybe later I'll try to display them better (a long vertical list of checkboxes might not be ideal).

package.json
jankeromnes: Yes, thanks.

server/konnectors/currencies.coffee
frankrousseau: Yes. A random id is created for each data.

server/konnectors/currencies.coffee
frankrousseau: Don't worry, now it's written in the manifest, NPM will do the job.

server/konnectors/currencies.coffee
frankrousseau: In fact I don't know if we handle checkbox at all on the client side. That's something that we will probably have to had.

server/konnectors/currencies.coffee
jankeromnes: Thanks! I'll have a look.

server/konnectors/currencies.coffee
jankeromnes: Does that include the import frequency?

server/konnectors/currencies.coffee
jankeromnes: I think the front-end will produce a `<input type="checkbox">`, which could work. If not I'll  make `"checkbox"` a special case.

server/konnectors/currencies.coffee
jankeromnes: I'm not sure how `query` and the `map` function interact...

How would I make a request that gives me the `CurrencyRate` where `currency="USD"` and `date` is the latest available? Also the request can be limited to `date` between today and 90 days ago, because we don't care about older dates anyway.

frankrousseau:
Ok. I will perform a better review later. Don't change the history, it will make the current discussion hard to read.

server/konnectors/currencies.coffee
frankrousseau: I'm not sure to understand, the comment `older  than 90 days` is confusing for me.

jankeromnes: It's a comment because I didn't know how to properly check `is date older than 90 days?`, but I've found example uses of `moment()` so I'm figuring it out right now. No need to worry about that.

server/konnectors/currencies.coffee
jankeromnes: Figured it out.

server/konnectors/currencies.coffee
frankrousseau: I think it matches with the fields property. So it's ok.

server/konnectors/currencies.coffee
frankrousseau: +1 `moment` makes things clearer.

server/konnectors/currencies.coffee
jankeromnes: Removed.

server/konnectors/currencies.coffee
jankeromnes: I switched to using a `CURRENCIES` array of supported codes just to be safe.

server/konnectors/currencies.coffee
jankeromnes: Done. I assume `moment() < moment()` comparisons work as expected.

server/konnectors/currencies.coffee
jankeromnes: Fixed.

server/konnectors/currencies.coffee
jankeromnes: Good to know, thanks! (But I guess `key`, `startkey` and `endkey` in requests still refer to the `date` field and not the random ID)

jankeromnes:
I won't change history until the code is ready to be merged. If at that time, you still wish history to be preserved for this pull request, I'll close it as is and open a new pull request from a different branch with all my commits squashed.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+            if moment(date) < moment(yesterday)
+                # Older than yesterday, refetch the last 90 days.
+                url = recentUrl
+
+        options =
+            method: 'GET'
+            url: url
+
+        request options, (err, res, body) ->
+            if err return callback err
```
jankeromnes: Ici, Travis se plaint d'un `unexpected if` mais je comprend pas pourquoi.

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+# Konnector
+
+module.exports =
+
+    name: 'Currencies'
+    slug: 'currencies'
+    description: 'konnector description currencies'
+    vendorLink: 'https://www.ecb.europa.eu/'
+
+    fields: # TODO Can we use the `CURRENCIES` array here?
```
frankrousseau: Yep what about:

```coffeescript
# Build fields: we want one checkbox for each available currencies.
fields = {}
fields[currency] = 'checkbox' for currency in CURRENCIES

# Konnector

module.exports =

    name: 'Currencies'
    slug: 'currencies'
    description: 'konnector description currencies'
    vendorLink: 'https://www.ecb.europa.eu/'

    fields: fields
```

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+        SGD: 'checkbox'
+        THB: 'checkbox'
+        ZAR: 'checkbox'
+    models:
+        currencyrate: CurrencyRate
+
+    # Define model requests.
+    init: (callback) ->
+        map = (doc) -> emit doc.date, doc
+        CurrencyRate.defineRequest 'byDate', map, (err) ->
```
frankrousseau: If you don't do anything before calling the callback, you can write it:

```
CurrencyRate.defineRequest 'byDate', map, callback
```

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+        # TODO Compute the last recent (max 90 days old) rates we know.
+        lastRates = {}
+
+        yesterday = moment().subtract('days', 1).format('YYYY-MM-DD')
+
+        # Figure out how far back in time we need to fetch.
+        url = dailyUrl
+        for currency of CURRENCIES
+            if !requiredFields[currency] #.checked ?
```
frankrousseau: It's more coffeescript to write it that way:

```coffeescript
if not requiredFields[currency]
```

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+                # We don't care about this currency.
+                continue
```
frankrousseau: I prefer one indentation to continue (I know it's matter of taste).

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+            date = lastRates[currency].date
+            if !date
+                # Jackpot! We have to refetch everything from the beginning...
+                url = histUrl
```
frankrousseau: The comment here is unclear.

server/konnectors/currencies.coffee
frankrousseau: The way you write it is not the coffeescript way. Here is the right line:

```coffeescript
return callback err if err
```

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+        request options, (err, res, body) ->
+            if err return callback err
+
+            xml2js.parseString body, (err, result) ->
+                if err return callback err
+
+                days = result['gesmes:Envelope'].Cube[0].Cube
+
+                for day of days
```
frankrousseau: you should use:

```coffeescript
for day in days
```

server/konnectors/currencies.coffee
```diff
@@ -0,0 +1,157 @@
...
+
+        yesterday = moment().subtract('days', 1).format('YYYY-MM-DD')
+
+        # Figure out how far back in time we need to fetch.
+        url = dailyUrl
+        for currency of CURRENCIES
+            if !requiredFields[currency] #.checked ?
+                # We don't care about this currency.
+                continue
+            date = lastRates[currency].date
```
frankrousseau: Because lastRates is an empty object, this will always raise an exception.

server/konnectors/currencies.coffee
jankeromnes: Right, thanks! Fixed in 2 places.

server/konnectors/currencies.coffee
jankeromnes: Fixed in several places.

server/konnectors/currencies.coffee
jankeromnes: I wasn't completely sure about what you meant, but now I use `continue if not requiredFields[currency]` with one indentation inside the for loop.

server/konnectors/currencies.coffee
jankeromnes: Rephrased to make it clearer.

server/konnectors/currencies.coffee
jankeromnes: Fixed in several occurrences, JS made me wary of using `in`.

server/konnectors/currencies.coffee
jankeromnes: I hope `?.` will do the trick.

jankeromnes:
Travis seems to choke on a few tests.

server/konnectors/currencies.coffee
jankeromnes: I'm still not sure how to do that request properly. It's not necessary for the Konnector to be functional, but it would be nice not to refetch all of history on every sync (even though it's only about 70 kB). I'll spend more time on this later.

jankeromnes:
I've renamed the Konnector to "Forex" because it's shorter.
I've also made it clearer that rates are based on the Euro (which shouldn't show up in the list).
I've also installed `brunch` (which should probably be a dev-dependency) to `brunch build` the client. This seems to have fixed the tests on Travis.

The Konnector fields now look like:
forex

However, when I ask to sync a few currencies, it quickly says "Imported" but the checkboxes are reset. How can I figure out if data was really imported?

jankeromnes:
The problem probably stems from the fact that `value` doesn't make sense for an `<input type=checkbox>` (for some reason it seems to always be `"on"`, regardless if checked or not). Instead, the `checked` attribute should be used.

frankrousseau:
Ok. Does the import works?

server/konnectors/forex.coffee
```diff
@@ -0,0 +1,127 @@
...
+                        last = lastRates[currency]
+                        if last and date <= moment(last.date)
+                            # We already know this rate.
+                            # TODO log.error if rate !== last.rate
+                            continue
+                        currencyrate = new CurrencyRate
+                            date: date
+                            rate: rate
+                            currency: currency
+                            base: 'EUR'
```
frankrousseau: Shouldn't that one be customisable?

jankeromnes:
I don't know if the import works, how can I verify?

server/konnectors/forex.coffee
jankeromnes: The exchange rates provided by the European Central Bank are all based on the Euro (which makes sense) but the conversion is quite trivial to do, so I let the consumers of the data handle that (e.g. in Kresus you will be able to switch between any base currency, using this data).

frankrousseau:
Do you use our development Virtual Machine?

If yes, you can run your connector then see the result by connecting to your CouchDB WebUI on `http://localhost:5984`.

If not, have a look at this tutorial: http://cozy.io/hack/getting-started/setup-environment.html

server/konnectors/forex.coffee
frankrousseau: Ok.

frankrousseau:
I tried it. The only things that doesn't work are the checkboxes: everything is fetched everytime. I was thinking: is that really a problem if everything is fetched every time?

jankeromnes:
Thanks for trying! I use Cozy in a VirtualBox (still waiting for my CubieTruck to arrive), but I don't seem to have anything running on port 5984, I'll have to check.

I also fixed the checkbox input problem (handling checkboxes differently by replacing `value="#{value}"` by a `#{checked}` attribute, and using `input.prop('checked')` instead of `.val()`). I'll push that commit here soon.

The main reason why I try not to fetch everything every time is to avoid overloading the ECB servers, but that optimization may be overkill. Still, I noticed that for historical values (from 2000 or so) the server sometimes took a little longer to reply, so it does help somewhat.

jankeromnes:
The latest commit fixes the checkbox issues, and adds a notification for imports. However, import seems to continue forever. Where can I see the logs to investigate what happens during import?

jankeromnes:
Also, my Konnector doesn't respect `requiredFields['date']` if a "starting date" was set. Maybe it should?

frankrousseau:
Ok, for development purpose it's better to use our development environment. 
Here is the link againt to set it up (it's based on Vagrant): http://cozy.io/hack/getting-started/setup-environment.html

The starting date is only here to allow the user to say when he wants the data fetching occurs. Don't worry about it.

About log you can display them by running:

    cozy-monitor log konnectors cat

or 

    cozy-monitor log konnectors tail

jankeromnes:
Thanks for all the excellent feedback! I've folded all suggestions and changes into a new branch, leading to pull request #51. Closing this one.

PR #51, opened 2015-04-26
Add Forex Konnector (new)

jankeromnes:
This pull request is based on #44, but it now saves a day's multiple rates into one single CouchDB document.

@frankrousseau please have a look!

A few questions:

1. I left a `# TODO` to "compute the latest recent rates (max 90 days old) that we know". How can I do this? I don't know how americano requests work, please help.
2. My code works for importing new rates. However, if I want to add a tracked currency, and re-do the import, old entries will not be updated, but duplicated. How can I fix this?
3. Should I refactor my code to use `fetcher.new()` and several `.use()` filters? If so, which ones?
4. Does my `brunch build` client make sense in this pull request? It's just a rebuild of the client.


server/konnectors/forex.coffee
```diff
@@ -0,0 +1,135 @@
...
+
+dailyUrl = 'https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml'
+recentUrl = 'http://www.ecb.europa.eu/stats/eurofxref/eurofxref-hist-90d.xml'
+histUrl = 'https://www.ecb.europa.eu/stats/eurofxref/eurofxref-hist.xml'
+
+# Other urls
+# http://www.global-view.com/forex-trading-tools/forex-history/index.html
+# http://www.global-view.com/forex-trading-tools/forex-history/exchange_csv_report.html?CLOSE_1=ON&start_date=1/1/1900&stop_date=03/06/2015&Submit=Get%20Daily%20Stats
+# http://www.global-view.com/forex-trading-tools/forex-history/exchange_csv_report.html?CLOSE_1=ON&CLOSE_2=ON&CLOSE_4=ON&start_date=1/1/1900&stop_date=03/06/2015&Submit=Get%20Daily%20Stats
+# http://www.federalreserve.gov/datadownload/Output.aspx?rel=H10&series=a6a0113179fdeb9c6fbcdc18575ec09c&lastObs=&from=01/01/2000&to=03/07/2015&filetype=csv&label=omit&layout=seriescolumn
```
jankeromnes: Note to self: Remove these URLs.

server/konnectors/forex.coffee
```diff
@@ -0,0 +1,135 @@
...
+    # Define model requests.
+    init: (callback) ->
+        map = (doc) -> emit doc.date, doc
+        CurrencyRate.defineRequest 'byDate', map, callback
+
+    # Fetch rates data from the ECB website and save them as Cozy objects.
+    fetch: (requiredFields, callback) ->
+        log.info 'import started'
+
+        # TODO Compute the latest recent rates (max 90 days old) that we know.
```
jankeromnes: Please help me with this!

server/konnectors/forex.coffee
```diff
@@ -0,0 +1,135 @@
...
+# http://www.global-view.com/forex-trading-tools/forex-history/index.html
+# http://www.global-view.com/forex-trading-tools/forex-history/exchange_csv_report.html?CLOSE_1=ON&start_date=1/1/1900&stop_date=03/06/2015&Submit=Get%20Daily%20Stats
+# http://www.global-view.com/forex-trading-tools/forex-history/exchange_csv_report.html?CLOSE_1=ON&CLOSE_2=ON&CLOSE_4=ON&start_date=1/1/1900&stop_date=03/06/2015&Submit=Get%20Daily%20Stats
+# http://www.federalreserve.gov/datadownload/Output.aspx?rel=H10&series=a6a0113179fdeb9c6fbcdc18575ec09c&lastObs=&from=01/01/2000&to=03/07/2015&filetype=csv&label=omit&layout=seriescolumn
+
+# Models
+
+CurrencyRate = americano.getModel 'CurrencyRate',
+    date: Date
+    rates: (x) -> x
```
jankeromnes: Note: Here, `(x) -> x` is not an array, but a map of currency code to rate (e.g. `USD -> 1.12`). I thought this was better than an array of pairs (e.g. `0 -> { currency: "USD", rate: 1.12 }`).

frankrousseau:
Thank you @jankeromnes for your PR. I will have a look at it and answer to your question on Thursday or on Friday.

frankrousseau:
>  I left a # TODO to "compute the latest recent rates (max 90 days old) that we know". How can I do this? I don't know how americano requests work, please help.

You have to declare your requests before starting the application. That's why you have to write them in the `server/models/request.coffee` file. They are CouchDB requests, the principle is that you run a map and a reduce functions on the whole set of documents with the right type to create a big hash. Then you claim values via keys. Do you know about CouchDB request or does it require an explanation?

>  My code works for importing new rates. However, if I want to add a tracked currency, and re-do the import, old entries will not be updated, but duplicated. How can I fix this?

See above. You have to implement the proper requests to know if the entry already exists.

>  Should I refactor my code to use fetcher.new() and several .use() filters? If so, which ones?

From my point of view, this archicture is proper for Konnectors. But your konnector is special, so I don't think you will be able to use existing "operators".

> Does my brunch build client make sense in this pull request? It's just a rebuild of the client.

Brunch build is not required, I prefer if you don't add it and handle this step myself.

frankrousseau:
I close this PR due to inactivity. Sorry @jankeromnes.

Issue #52, opened 2015-05-01
Add file name for bank operation


frankrousseau:
When a link is created between a bank operation and a bill binary, there is only a link to the binary. But it would be nice to have the original file name.

bnjbvr:
So that's the right way to solve https://github.com/bnjbvr/kresus/issues/97 ?

PR #72, opened 2015-05-24
Save file name and file mime for clean link to bankoperation

ZeHiro:
Fixes https://github.com/cozy-labs/konnectors/issues/52

This PR allows also to update already linked operations.

One question remains (just editorial) : 
do I rather use another name for fileName and fileMime ? (using only lower case field names)

### DIFF
--- a/server/lib/link_bank_operation.coffee
+++ b/server/lib/link_bank_operation.coffee
@@ -50,8 +50,8 @@ class BankOperationLinker
             amount = parseFloat entry.amount
         catch
             amount = 0
-
         for operation in operations
+        
             operationAmount = operation.amount
             if operationAmount < 0
                 operationAmount = operationAmount * -1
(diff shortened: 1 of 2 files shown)


frankrousseau:
Thank you for working on this. But the binary object is an agnostic object. I think it's better if you save this at the bill level only. 
Another problem will be if the user changes the file name. It may happen rarely, I'm not sure we should handle it.

ZeHiro:
OK. I'll rty to propose something else. A simple question, after a fast review of all the "bill import" konnectors, it appears PhoneBill and InternetBill have the same structure (codeBill has one more field), is there a goal for that? Couldn't there only be a "bill" structure? My point is that I plan to create a connector to import bills from Capitaine Train, so train bills. But I don't see the interest of creating a new specific structure (e.g. TrainBill) just for that. I would create a more generic structure.

frankrousseau:
Sorry I just realised I never answered to your question. I totally agree that a Bill object with a type field would be better.

ZeHiro:
OK. I'll see wht I can do for that. But a transition has to be done.

frankrousseau:
Let's merge it and not worry about file name changing. Whatever the orginal name is better than no name.

frankrousseau:
Thank you @ZeHiro

PR #73, opened 2015-05-25
Prevent starting a timer twice

clochix:
This is a backport of a fix that seems to work on ISEN branch.


frankrousseau:
Thank you @clochix for the fix.

Issue #75, opened 2015-06-10
Free konnector broken


ZeHiro:
https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/free.coffee#L102
and L103
These lines should be : 
        isError = res.headers.location? and
            res.headers.location.indexOf("error") isnt -1

ZeHiro:
I'll propose a PR for this issue and for a fix for freemobile connector

frankrousseau:
Ok thank you. It's probably a mistake I did in my last merge.

PR #76, opened 2015-06-10
Fixes for free and freemobile konnectors 

ZeHiro:
Fix for  https://github.com/cozy-labs/konnectors/issues/75

and fix of bad credential management for freemobile konector
Fix timer for freemobile konnector: I topped it to 14 days (I had an unlinked operation). 14 is chosen to be also applicable for februray (28days) and avoid mismatch for this month


frankrousseau:
OK thank you!

PR #79, opened 2015-06-23
Fix github commit and add build

poupotte:


clochix:
Thanks @poupotte, I asked @eravion and he confirmed that he was able to install the application on its deb based Cozy.
One the application installed, he founds a lot of bugs in Konnectors, expect tickets ;-)

PR #84, opened 2015-07-01
Add nest konnector

cylwin:


frankrousseau:
As discussed privately I close that one. I integrated the Nest konnector myself.

PR #89, opened 2015-07-06
FInal migration to cozydb

ZeHiro:


frankrousseau:
Arg, thank you... I forgot si many things during this migration...

Issue #97, opened 2015-07-15
[Free Mobile] Manage the case where the session is already established


ZeHiro:
When 2 imports are done in a short periode of time without restarting the application the import fails ("wrong credentials" error raised) because the session is still active.  (session cookie is kept), the keyboard is not proposed.

PR #98, opened 2015-07-15
Fix logout of Free Mobile konnector

ZeHiro:
Fixes https://github.com/cozy-labs/konnectors/issues/97


frankrousseau:
Thx for the fix!

PR #100, opened 2015-07-20
UI: added a default icon + digital ocean icon + sosh icon

kelukelu:
I couldn't try it on my local version (Even with a pull, I didn't see Digital Ocean or Sosh).
But it should be fine :)


frankrousseau:
Great! Thank you @kelukelu!

Issue #92, opened 2015-07-10
[Free Mobile]  Make the difference between sevel phone numbers for the same account


ZeHiro:
Support multi line (several phone numbers for one account). Use the "phone numbre" to make the difference.

f6fvy:
I'm in that use case, and currently there a 2 different pdf files imported : One for the detailed phone bill of the "master number" of the account, and another one which is the summary of the various attached phones numbers. **But** these two files have the same name in the files app, which is kind of disturbing !

ZeHiro:
OK. I will make an update of the konnector, I wait for some other fixes to implement it in a clean way. Can you describe me what files are listed on the webpage? And if it is necessary to import all of them? I only have one line attached to my free mobile account. 
When the update is made, could you then test-it?

frankrousseau:
Thank you for reporting.

PR #115, opened 2015-08-04
Fix #112 , #92 plus minor fix in save_data_and_file librairy

### LINKED ISSUE #112: [Free Mobile] Fix pdf URL
Currently the pdf URL matched leades to an html page. 
The url should be updated to be "built" and not only get from a link. 

Question: is it possible to force reimport a file without touching databrowser? Should be necesseary. Maybe it could be interessting to check if the binary attached to the Bill (or Phone/Internet...bill) exist. If it does not, reimport

ZeHiro:


frankrousseau:
Awesome thank you @ZeHiro

PR #120, opened 2015-08-07
Import private github commits

nono:
Currently, only public commits are imported. With this PR, private commits are also imported.


frankrousseau:
Thank you!

PR #132, opened 2015-08-12
Numericable

### LINKED ISSUE #56: Numericable bills
Another Konnector idea.

nicofrand:
That should fix #56.

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -67,6 +67,7 @@ module.exports =
     'konnector description sosh': "Download all your Sosh Bills. This konnector requires the Files application to store the bill PDF files."
     'konnector description electrabel': "Download all you Electrabel Bills. This konnector requires the Files application to store the bill PDF files."
     'konnector description orange': "Download all your Orange Bills. This konnector requires the Files application to store the bill PDF files."
+    'konnector description numericable': "Download all your Numéricable Bills. This konnector requires the Files application to store the bill PDF files."
 
     # Konnectors' notifications
     'notification prefix': "Konnector %{name}:"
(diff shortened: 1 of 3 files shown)


frankrousseau:
Thank you for this new konnector! Before merging it, it would be better if the code follows the Cozy guidelines: 
https://github.com/cozy/cozy-guidelines/#cozy-code-guidelines

What is missing: 

* replace `!` by `not` keyword.
* Lines should be < 80 chars. 

Can you handle it?

nicofrand:
Yes I will.
Since you read it, could you tell me what was wrong with the following code please ?

```
Bill = cozydb.getModel 'Bill',
    vendor: {type: String, default: 'Internet'}
    type:
        type: String
        default: 'Internet'
    vendor:
        type: String
        default: 'Numéricable'
     date: Date
     amount: Number
```

I had to replace it with:
```
Bill = require '../models/bill'
```

I followed https://github.com/cozy-labs/konnectors/wiki#data-description-models

frankrousseau:
I don't know what's wrong except the fact you declared the vendor field twice. Indentation doesn't look proper too.
But the real reason is probably that the Bill model is already declared in the model folder.

nicofrand:
The indentation is only wrong because I copy/pasted this block wrong.
I am pretty sure I tried the vendor things twice too and just forgot to remove one of them once commented.

Why does the wiki says to use getModel if the model already declares it  then ?
(And by the way, why is there "vendor: type: String, default: 'Digital Ocean'" ? Digital Ocean should not be the default vendor)

frankrousseau:
Great! Thank you @nicofrand for this konnector.
Setting vendor must be done for each bill. But the save_data_and_file layer do it for you.

About the Digital Ocean as a vendor it's a mistake that will be fixed soon.

Thank you again for the bug reports.

nicofrand:
Where do you see that save_data_and_file already sets the vendor automatically ?

When it does, does it use the third parameter:
`.use(saveDataAndFile(log, Bill, 'numericable', ['bill']))`

I suggest to complete the wiki with those information.

ZeHiro:
@nicofrand , the third parameter will set the vendor automatically, if it is not already set in the bill object you created when you parsed them form the website.

nicofrand:
Thanks @ZeHiro , I updated the wiki so that is it clear now.

PR #154, opened 2015-10-14
Capitalized konnector name.

babolivier:
ISEN is an acronym, I think it's better to have all the name capitalized.


babolivier:
(sorry for additionnal commits, I messed up between my branches)

PR #171, opened 2015-12-21
Fix import orange

Peltoche:
Add check on request response

### DIFF
--- a/server/konnectors/orange.coffee
+++ b/server/konnectors/orange.coffee
@@ -106,7 +106,7 @@ logIn = (requiredFields, billInfos, data, next) ->
                 log.info 'Login succeeded'
 
                 response = JSON.parse body
-                if response.credential? response.password?
+                if response.credential? and response.password?
                     error = if response.credential? then response.credential
                     else response.password
                     next new Error(error)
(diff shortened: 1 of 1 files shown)


server/konnectors/orange.coffee
```diff
@@ -105,17 +105,24 @@ logIn = (requiredFields, billInfos, data, next) ->
...
-                        log.error 'An error occured while fetching bills'
-                        console.log err
-                        next err
-                    else
-                        log.info 'Fetch bill info succeeded'
-                        data.html = body
-                        next()
+                response = JSON.parse body
+                if response.credential isnt undefined or
+                response.password isnt undefined
```
frankrousseau: You should use `if response.credential? and response.password`

server/konnectors/orange.coffee
```diff
@@ -105,17 +105,23 @@ logIn = (requiredFields, billInfos, data, next) ->
...
-                    if err
-                        log.error 'An error occured while fetching bills'
-                        console.log err
-                        next err
-                    else
-                        log.info 'Fetch bill info succeeded'
-                        data.html = body
-                        next()
+                response = JSON.parse body
+                if response.credential? response.password?
```
frankrousseau: A `and` keyword is missing here. Am I wrong or is it missing?

PR #174, opened 2016-01-07
Es6 konnector

Peltoche:
- Add build tools
- Refactor linkedin to ES6
- Move linkedin helper in extern librairy


server/konnectors/linkedin.js
```diff
@@ -0,0 +1,512 @@
...
+  },
+
+  models: {
+    contact: Contact
+  },
+
+  fetch: (requiredFields, callback) => {
+    log.info('Import started');
+
+    Promise.resolve(requiredFields)
```
frankrousseau: Could you wrap this in a fetcher object or function?

server/konnectors/linkedin.js
```diff
@@ -0,0 +1,512 @@
...
+
+/**
+* Load landing page to retrieve the csrf token needed in login request.
+* More info on csrf -> https://en.wikipedia.org/wiki/Cross-site_request_forgery
+*
+* The html parsing is done with cheerio, a "jquery like" that can be run on
+* the server side.
+*/
+function retrieveTokens(requiredFields) {
+  return new Promise((resolve, reject) => {
```
frankrousseau: Could you wrap this so the use doesn't have to write Promise in his code?

package.json
```diff
@@ -33,16 +33,16 @@
     "imagemagick-stream": "2.0.0"
   },
   "devDependencies": {
+    "cli-table": "^0.3.1",
     "coffee-script": "1.9.3",
     "coffeelint": "^1.9.4",
     "cozy-fixtures": "1.1.3",
+    "express": "4.13.3",
+    "minimist": "^1.1.3",
```
frankrousseau: could you remove \^in dependencies?

server/konnectors/linkedin.js
frankrousseau: Sorry I meant the konnector developer, not the user.

When I read this code: https://github.com/pouchdb/pouchdb/blob/master/tests/mapreduce/test.persisted.js#L57-L108

I see that it's possible to chain many functions without calling the Promise constructor. So do you think it's possible to remove the `new Promise` in your function?

Peltoche: I guess the pouchdb create a promise and return it so we can link directly to it the other promises. 
In all case when you haves a async function you need to create a promise, so in our case we have two solutions: we create ourself the promises (like this code) or we use a lib which manage himself the promise creation and return it (like your exemple).
Almost all async action are created by the request module and I juste see that he have an official promise version. So yes I probably can almost all "new Promise" terms but if you want remove all I need to move them in the Models or better in the CozyModel module.

frankrousseau:
I close this PR. Most the changes has been included in the my last push to development.

PR #176, opened 2016-01-13
Correct UTC format issue

Peltoche:
That could lead to not find the factures to the bank operations

### DIFF
--- a/server/konnectors/isen.coffee
+++ b/server/konnectors/isen.coffee
@@ -494,7 +494,7 @@ module.exports =
 
     createFile: (fileName, path, date, url, tags, callback) ->
         @numItems++
-        File.createNew fileName, path, date, url, tags, (err) ->
+        File.createNew fileName, path, url, tags, (err) ->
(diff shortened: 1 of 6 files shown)


Peltoche:
Hey @ZeHiro !

This PR mean to resolve this issue.
Before the use of moment().toISOString() lead to move the local time (UTC+1) to ISO format (UTC+0) and the issue you have explained. Now we don't use this function but parse the date manually to ISO and force it set the date to (YYYY-MM-DD)T00:00:00.000Z so they can match with the Kresus transaction.

Thanks for the two links, I wondered were else this error should be inserted. Now all should be good.

Peltoche:
The test are ok locally, I don't know what can fail in the konnector system.... @frankrousseau an y idea?

ZeHiro:
@Peltoche , good news :)

frankrousseau:
Don't worry about tests, I will check what's wrong later. Thx to both of you!

PR #187, opened 2016-02-09
Fix bouygues telecom konnector (urls and date parsing)

Chocobozzz:
It seems they changed their URL. And the date is now "Février 2016" for example so we need to set the french locale to moment.


frankrousseau:
Thank you @Chocobozzz

PR #195, opened 2016-02-11
Fix build on node < 4

nicofrand:
Brunch 2+ only works with Node 4+. 

I tried to downgrade brunch to version 1, everything seems OK regarding the build process.


frankrousseau:
Cozy is now made for Node 4. Why do you want to make the dev environment compatible with previous version of Node?

PR #199, opened 2016-02-18
Do not install client npm modules on production

nono:
It's 22M less to download and have on disk


frankrousseau:
Great, thank you!

Issue #54, opened 2015-05-01
OVH Bills


frankrousseau:
A konnector to download OVH bills.

nicofrand:
It looks like there is an API for that: https://api.ovh.com/console/#/me/bill#GET

frankrousseau:
I didn't see that! Thank you.

Chocobozzz:
WIP

frankrousseau:
Wow

brouberol:
Can this issue be closed?

PR #200, opened 2016-02-25
Add OVH konnector

Chocobozzz:
Feel free to change appKey/appSecret, these are just for tests.

This konnector is only for ovh-eu but there are other available services:

 * OVH Europe: ovh-eu (actual)
 * OVH North-America: ovh-ca
 * RunAbove: runabove-ca
 * SoYouStart Europe: soyoustart-eu
 * SoYouStart North-America: soyoustart-ca
 * Kimsufi Europe: kimsufi-eu
 * Kimsufi North-America: kimsufi-ca

We choose the service when we generate the login url. So IMHO there are two options:
 * Allow the user to choose the service with a select box in a "OVH global konnector". The drawback is they won't be able to have multiple simultaneous services (ovh-eu and kimsufi-eu for example).
 * Create a "global ovh backend" and one konnector per service (so 7 konnectors). The drawback is duplication of code (boring code like translations etc)

I personnaly prefer the second option to keep the "1 konnector for 1 service" philosophy.

Thoughts?

### DIFF
--- a/server/lib/base_ovh_konnector.coffee
+++ b/server/lib/base_ovh_konnector.coffee
@@ -1,0 +1,47 @@
+ovhFetcher = require '../lib/ovh_fetcher'
+filterExisting = require '../lib/filter_existing'
+saveDataAndFile = require '../lib/save_data_and_file'
+linkBankOperation = require '../lib/link_bank_operation'
+baseKonnector = require '../lib/base_konnector'
+
+Bill = require '../models/bill'
+
+module.exports =
+    createNew: (ovhApi, name, slug) ->
+
+        fileOptions =
+            vendor: slug
+            dateFormat: 'YYYYMMDD'
+
+        fakeLogger =
+            info: (text) -> connector?.logger.info(text)
+            error: (text) -> connector?.logger.error(text)
+            warn: (text) -> connector?.logger.warn(text)
+            debug: (text) -> connector?.logger.debug(text)
+
+        ovhFetcherInstance = ovhFetcher.new(ovhApi, slug, fakeLogger)
+
+        fetchBills = (requiredFields, entries, body, next) ->
+            ovhFetcherInstance.fetchBills(requiredFields, entries, body, next)
+
+        return connector = baseKonnector.createNew
+            name: name
+
+            fields:
+                loginUrl: "link"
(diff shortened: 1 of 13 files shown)


frankrousseau:
Thats sound good. I prefer the second option.

Thank you a lot for this konnector. And sorry for the late feedback. I was sick this week-end and on monday. So I wasn't able to look at it. I will make first in-depth review on friday or on next monday.

Chocobozzz:
Hi Frank

Sorry to hear that, hope you're feeling better. And no worries I'm in no hurry :)

So if we're both in the second option, I can implement it and squash the new commits into this PR.

Chocobozzz:
Done :)

I just tested OVH EU because I don't have any account on the other services.

kosssi:
Fix https://github.com/cozy-labs/konnectors/issues/54

server/lib/base_ovh_konnector.coffee
```diff
@@ -0,0 +1,47 @@
...
+            fields:
+                loginUrl: "link"
+                token: "hidden"
+                folderPath: "folder"
+
+            models: [Bill],
+
+            fetchOperations: [
+              fetchBills,
+              filterExisting(fakeLogger, Bill)
```
frankrousseau: Why do you use a fakeLogger instead of the `konnector.logger`?

Chocobozzz: Because at this point unfortunately `connector` is undefined

frankrousseau: Sad.... I don't know how to make it proper.

frankrousseau:
I can now have my OVH bills automatically fetched in my Cozy!

Chocobozzz:
Just miss the Application keys/secrets for each service (https://github.com/ovh/node-ovh/#1-create-an-application) :)

frankrousseau:
Arg that sounds bad. Maybe we should rewrite it to just scraps the data?

Chocobozzz:
Why? It's just a form and then they give you your Application credentials.

frankrousseau:
It's one app per user or one app for all users?

Chocobozzz:
One app for all users.

PR #202, opened 2016-03-06
Add Ical feed connector

frankrousseau:
Add a connector to fetch data from an Ical feed on a regular basis.

It comes with a base object for connectors to make the development of connectors easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "konnectors",
-  "version": "0.4.8",
+  "version": "0.4.11",
   "engines": [
     "node = 0.10.x"
   ],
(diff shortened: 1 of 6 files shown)


Chocobozzz:
Awesome :+1: 

The base konnector will simplify a lot some konnectors.

Just have a question: why do you use `var` instead of `let`?

frankrousseau:
No reason :). What `let` does better than `var`?

Chocobozzz:
I'm not an ES6 expert but I read some posts saying that we should use `let` instead of `var` with ES6.

This answer on stackexchange provides a good argumentation in my opinion: http://programmers.stackexchange.com/a/274352

frankrousseau:
Ok, thanks for the tip!

Issue #205, opened 2016-03-08
Konnector for voyages-sncf.com


Chocobozzz:
To add bills, tickets and calendar events in Cozy.

Chocobozzz:
WIP

PR #222, opened 2016-04-02
Add SNCF konnector

Chocobozzz:
Add events and bills (without pdf) in Cozy. So we can see travels details in the calendar app :)

I'm sorry in advance, the code is ugly because the generated HTML by SNCF is a little bit difficult to parse (#frenchtech). Open to remarks.

Closes https://github.com/cozy-labs/konnectors/issues/205

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -110,6 +110,7 @@ module.exports =
     'konnector description googlecontacts': "Import your google contacts into your Cozy through google's API."
     'konnector description linkedin': "Import your Linkedin contacts in your Cozy."
     'konnector description ameli': "Import your Ameli reimbursements in your Cozy."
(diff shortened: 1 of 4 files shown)


frankrousseau:
Thank you for that new konnector. I will be super glad to use it. It doesn't fetch the history. Is it something you could add?

Chocobozzz:
Yep, but it would be just for the bills. I keep you in touch :)

frankrousseau:
Another thing. Because we don't have the PDF file. Can you add a content field to the bill model and save details into it?

Chocobozzz:
I pushed the modifications. Waiting for https://github.com/cozy-labs/konnectors/pull/223 before reformating.

Chocobozzz:
Done.

frankrousseau:
Thank you @Chocobozzz!

Issue #216, opened 2016-03-19
Connection to ameli fail


encolpe:
From the log:

```
[Fri, 18 Mar 2016 23:57:42 GMT] - PUT /konnectors/a87eb666a966bccd4b389963e9a5f1a1 - 200 - 54.395 ms -  16
[2016-03-19 00:57:42:977] info - ameli : Next update 26/03/2016 at 00:57:42

/usr/local/cozy/apps/konnectors/build/server/konnectors/ameli.js:59
        'Cookie': res.headers['set-cookie'],
                     ^
TypeError: Cannot read property 'headers' of undefined
    at Request._callback (/usr/local/cozy/apps/konnectors/build/server/konnectors/ameli.js:59:22)
    at self.callback (/usr/local/cozy/apps/konnectors/node_modules/request/request.js:198:22)
    at Request.emit (events.js:95:17)
    at Request.onRequestError (/usr/local/cozy/apps/konnectors/node_modules/request/request.js:820:8)
    at ClientRequest.emit (events.js:95:17)
    at CleartextStream.socketErrorListener (http.js:1547:9)
    at CleartextStream.emit (events.js:95:17)
    at SecurePair.<anonymous> (tls.js:1384:19)
    at SecurePair.emit (events.js:92:17)
    at SecurePair.maybeInitFinished (tls.js:977:10)
```

jeekajoo:
Same here. Let me know if you need more info.

frankrousseau:
ping @Gara64

frankrousseau:
Thank you @encolpe  and @jeekajoo for reporting.

Gara64:
It looks like the first GET request to ameli fails, but I cannot figure out why, and I can't reproduce.
@encolpe and @jeekajoo , do you have others konnectors working? 
And to be sure everything is correctly set up, could you uninstall the app then reinstall and retry?
By this time I'll handle this error to have more logs.
Thanks!

encolpe:
I have the same problem with LinkedIn, but Free, Online and Orange connectors are working.

jeekajoo:
all other konnectors I have work, uninstalling / reinstalling did not help.

jeekajoo:
with version 0.5.3: `An error occurred during the last import: CERT_UNTRUSTED`

Gara64:
[...]

PR #227, opened 2016-04-19
Fix certificate issue

Gara64:
As mention on #216, the last debian version considers the Ameli's certificate issuer as untrusted.
This should be a temporary patch, waiting for the [bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812708) to be resolved on debian.


frankrousseau:
Thank you @Gara64 !

PR #228, opened 2016-04-23
Sfrbox et sfr mobile

doubleface:
This works also for red clients and has been tested by some users : https://forum.cozy.io/t/konnectors-nouveau-connecteur-pour-sfr/2348


frankrousseau:
Hi @doubleface,

  Thank you for that set of new connectors. SFR customers will like it. About the code I have two remarks:

* Can you use the `Bill` model instead of `InternetBill` and set type field with *Internet* as value (on each bill)?
* Can you use make sure that there is no linting errors (`npm run lint`)?

doubleface:
OK, I will take à look at it

doubleface:
Bon, cette fois les tests n'échouent que à cause du module sncf. Donc à priori c'est bon.

doubleface:
Ooups, this time, the unit tests only fail because of the sncf module. This one is the good one, I suppose.

SomeAverageDev:
You will just need to add the locales for SFR MOBILE and you should be done for good !

frankrousseau:
Thank you for the modification. I have two other changes to ask you before merging:

* Can you use `base_konnector` to make the description of your konnector lighter?
* Can you require directly the model `..//models/bill` instead of redescribing it and defining its queries?

doubleface:
And it is done

frankrousseau:
Thank you a lot. It's now included in the last release of konnectors!

PR #229, opened 2016-04-23
Fix schedules and place details in SNCF konnector

Chocobozzz:


frankrousseau:
Thank you for the fixes! It will be included in the next 0.6.0 version.

PR #248, opened 2016-05-07
Fixed server-side first import date

doubleface:
Fixing problem exposed in https://forum.cozy.io/t/konnectors-nouveau-connecteur-pour-sfr/2348/30. The server read the date in english. There is some more work to do to handle the locale here.


frankrousseau:
Thank you for the fix @doubleface.

Issue #250, opened 2016-05-12
When an import does not provide new data at notification is still raised


ZeHiro:
It happends for ical connector and doctolib connector. Same with free.

doubleface:
And same with SFR or ameli. I Will look at it since this is probably due to my notification fix

ZeHiro:
@doubleface 
The problem is still here on manual fetch

I think it is here : 
https://github.com/cozy-labs/konnectors/blob/20912f900bf788dccc613cd71cbd4b22a9730098/server/controllers/konnectors.coffee#L94

doubleface:
Yes, you are right, I fixed this this week end but could not finish it because it raised another problem with localization. I may update the pullrequest later this week (if I get the time to do  it...).

ZeHiro:
@doubleface I took care of that

doubleface:
Thank you for that @ZeHiro !

PR #255, opened 2016-05-15
Fix issue #250 : do not raise empty notification when nothing appened

doubleface:
I could not test it on my instance but I suppose it will work.


frankrousseau:
Thx!

PR #258, opened 2016-05-16
Delete HTML balises of error message server side.

### LINKED ISSUE #256: Sanitize error message
Some website send html error message. They should be sanitized to prevent client failure.

ZeHiro:
Fixes #256


frankrousseau:
Thx again!

PR #266, opened 2016-05-22
Direct Energie konnector

bnjbvr:
This provides a new Konnector for Direct Energie, that can fetch bills on the web site as well as save PDFs when there are some. It can also link bills to bank transactions: rather than simply using the entry's date, I've used the "paid at date", which is a field that might or might be not proposed by the document provider. Direct Energie does that, and the date delta is rather small when there's a "paidAtDate" field. I hope some other konnectors can also make use of this paidAt field :-) Note it's not saved in the DB, just used for linking bank operations to files.

Please let me know if I can do anything to ease the review :)

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -121,7 +121,7 @@ module.exports =
     'konnector customview googlecontacts 1': "1. Press \"connect your google account\" button to connect to your Google account and authorize your Cozy to access to it. Google will provide you with a complex string. Once you get it copy it in your clipboard, we will use it in second step."
     'konnector customview googlecontacts 2': "Connect your Google account"
     'konnector customview googlecontacts 3': "2. Paste this string in the Auth code field. Then press 'Import and save' button to start the sync. Account name will be automatically updated."
-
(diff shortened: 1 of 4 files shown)


frankrousseau:
@bnjbvr everything looks good except the linting. Can you run `npm run lint` to ensure that it's ok?

frankrousseau:
(btw, thank you for this new connector :))

ZeHiro:
@bnjbvr,
could you add the case where credendials are wrong? based on the res.statusCode and the res.headers.location if existing, when detecting the credentials are wrong, you shoud also call next 'bad credentials' so that the user is warned the login failed

bnjbvr:
Thank you for the quick reviews! Wow, your linting rules are quite strict (I don't quite understand the need to enforce `const` rather than `let` for bindings that are not mutated; as a VM implementer, I'd think it's even slower to use `const` in this case, but anyways).

@ZeHiro, the website behaviors when the login isn't correct is to return the login page with a div containing the text error itself (the response status code is 200, sic). I've made a quick pattern matching to catch this, but it is rather fragile.

frankrousseau:
Thx @bnjbvr for the fixes!

Issue #213, opened 2016-03-15
Orange pdf match problem with Kresus


eravion:
Hello
2 bugs.

1/ Match with bad line
Info pdf : total : 66.69 / date :  11.12.2015
this line in Kresus match
montant :  69 / Date : 21.11.2015

This line have to match but is not 
montant :  66.69 / Date : 14.12.2015

2/ Does not match
Info pdf : total : 32.59 / Date : 12.03.2016
line in Kresus have to match
montant : 32.59 / Date : 14.03.2016

eravion:
Hi,

I still have the problem. Sorry

PR #268, opened 2016-05-22
Try a smarter operation / bill match algorithm;

bnjbvr:
I was looking at #213 and thought about a simple situation where the match would lead to the first situation: if we match the right operation first, but then there happens to be another operation which matches the criterias too, we'll end up clobbering the right operation to link. So I think the linking should try to minimize the difference in amount, at least, which is what this PR does.

Also: I've removed `minAmountDelta` and `maxAmountDelta`, as nobody were using those in the konnectors directory. It also makes the comparisons simpler.

Also changed the default value of `amountDelta` to 0.001, because IEEE754 and comparing two floating point values for equality is equivalent to comparing the bits. If you're not convinced, run this piece of code: `if (0.1 + 0.2 !== 0.3) alert("LET THE WORLD BURN");`. :-)

Also, there was no clear reasons why we'd take the absolute value of the amount for the operation's amount but not for the reference amount; I've updated this for consistency.

Finally, I think that if we can't parse the amount, we'd better just skip the matching, as we'll end up either with an incorrect amount or a NaN.

### DIFF
--- a/server/lib/link_bank_operation.coffee
+++ b/server/lib/link_bank_operation.coffee
@@ -16,9 +16,7 @@ class BankOperationLinker
         @log = options.log
         @model = options.model
         @identifier = options.identifier.toLowerCase()
-        @amountDelta = options.amountDelta or 0
-        @minAmountDelta = options.minAmountDelta or @amountDelta
-        @maxAmountDelta = options.maxAmountDelta or @amountDelta
+        @amountDelta = options.amountDelta or 0.001
         @dateDelta = options.dateDelta or 15
         @minDateDelta = options.minDateDelta or @dateDelta
         @maxDateDelta = options.maxDateDelta or @dateDelta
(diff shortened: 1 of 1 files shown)


bnjbvr:
And a final note: I think we should also try to minimize the date delta, but as this introduces a second criteria for minimization, this makes things much more complicated (there could be an operation with a more precise amount but not a more precise date, etc.).

frankrousseau:
Thx @bnjbvr. It looks great, but it requires you rebase your PR. Can you handle this?

bnjbvr:
Sure, here we go!

bnjbvr:
Out of curiosity, would you have a lot of examples of real-world bank operations / bills to test against this algorithm? That would be very nice, especially for integrating a second criteria (date delta) to minimize.

frankrousseau:
No... A good option would be to add a canvs for test cases to unit tests related to these algorithms. Then we could invite people to add test cases based on their real-life use cases.

ZeHiro:
, there can be quite a lot of time between billing date and operation date. For amount i didn't see diffrenced

PR #286, opened 2016-05-27
Konnector for captain train 

### LINKED ISSUE #178: Konnector for Captain Train

ZeHiro:
Fixes #178 
Buying tickets on Captain Train can lead to different operation labels: 'Capitain Train', 'Capitaine train' (for older operations), for Ouigo tickets bought on capitaine train, the operation is directly labelled 'Ouigo'. I then allowed to use an array as the operation identifier to allow to manage these situations. To maintain compatity with existing konnectors,  if the passed label is a string, it is put inside an array.

Capitain Train connector can also import refund bills, to be sure the bills are matched to the appropriate operation, in case on the same day, a train ticket is bought, and the same amount is refounded, the distinction was made between normal payments, and refunds. Refunds need do be declared as such, using isRefund attribute of the Bill object. 

Still known problem : if exacty the same operation is done on the same day (by twice the same ticket with same amount ), one of the two operations won't be linked to a bill, but the bill will be imported.

### DIFF
--- a/server/lib/link_bank_operation.coffee
+++ b/server/lib/link_bank_operation.coffee
@@ -15,7 +15,10 @@ class BankOperationLinker
     constructor: (options) ->
         @log = options.log
         @model = options.model
-        @identifier = options.identifier.toLowerCase()
+        if typeof(options.identifier) is 'string'
+            @identifier = [options.identifier.toLowerCase()]
+        else
+            @identifier = options.identifier.map((id) -> id.toLowerCase())
         @amountDelta = options.amountDelta or 0.001
         @dateDelta = options.dateDelta or 15
         @minDateDelta = options.minDateDelta or @dateDelta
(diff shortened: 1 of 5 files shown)


Chocobozzz:
IMHO, you should `return` when calling the callback on error. I think it is worth to activate [this lint rule](http://eslint.org/docs/rules/callback-return) in ESLint.

ZeHiro:
Makes sense.

server/lib/link_bank_operation.coffee
```diff
@@ -49,21 +52,32 @@ class BankOperationLinker
 
         try
             amount = Math.abs parseFloat entry.amount
+            # By default, an entry is an expense. If it is not, it should be
+            # declared as a refund: isRefund=true.
+            if not entry.isRefund? || not entry.isRefund
```
frankrousseau: You could write:

```coffeescript
amount *= -1 if  entry.isRefund? and entry.isRefund
```

frankrousseau:
Could you explain what you are doing with the pattern matching?

ZeHiro:
I'm not sure i understand the question :s

frankrousseau:
yes and tell us what the changes do.

ZeHiro:
I have updated the comment of the PR.

server/lib/link_bank_operation.coffee
ZeHiro: Yep

frankrousseau:
Thx @ZeHiro

PR #287, opened 2016-05-29
Improve notification management

### LINKED ISSUE #250: When an import does not provide new data at notification is still raised

ZeHiro:
Manage notifications at the same location: previsously they were managed (the same) in the importer and the konnector controller.

This will help future maintenance of this function (previsously #250 was fixed only for automatic fetch).
Fixes the fact that only one notification could be sent for each konnector. Now one notification is raised per account.
Fixes empty notifications on manual fetch (fixes #250)

### DIFF
--- a/server/lib/notification_handler.coffee
+++ b/server/lib/notification_handler.coffee
@@ -1,0 +1,46 @@
+konnectorHash = require '../lib/konnector_hash'
+localization = require '../lib/localization_manager'
+NotificationHelper = require 'cozy-notifications-helper'
+notification = new NotificationHelper 'konnectors'
+slugify = require 'cozy-slug'
+
+log = require('printit')
+    prefix: null
+    date: true
+
+# Create a notification telling how many data were imported.
+module.exports = (konnector, notifContents) ->
+    model = konnectorHash[konnector.slug]
+
+    prefix = localization.t 'notification prefix', name: model.name
+
+    for index of notifContents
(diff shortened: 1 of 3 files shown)


server/lib/notification_handler.coffee
```diff
@@ -0,0 +1,46 @@
...
+
+    prefix = localization.t 'notification prefix', name: model.name
+
+    for index of notifContents
+        notificationSlug = konnector.slug
+
+        # For each account of the konnector, generate a unique nofication slug
+        # based on credentials
+        for credential in konnector.accounts[index]
+            notificationSlug += "_#{slugify credential}"
```
frankrousseau: Are you sure it doesn't include passwords?

frankrousseau:
Thank you @ZeHiro for this fix. I only have a security question. I merge the PR but can you confirm that everything works fine?

ZeHiro:
@frankrousseau, password are in another field of the konnector doctype. But as the password are decrypted in the account (can easily be found just opening the console and changing the type of input from password to text) there is at least no decrease of security.
If the konnctor devlopper didn't use the password field but another one to store it, it will be used. 
 And yes it works.

ZeHiro:
After rereading the code, it appears the konnectors with passwords won't be stored correctly, it seems, but i am not sure, the passwords are not store in a array. To be confirmed

Issue #294, opened 2016-05-31
Match same operation on the same day


ZeHiro:
With capitain train, it is possible to have on the same day, the exact same operation, without being duplicates (just buy twice the same ticket, but pay it in two time). Currently, it is possible to match only one operation, the second one won't be.

PR #295, opened 2016-05-31
fix matching amount

ZeHiro:
@frankrousseau 
This fix is good and tested. Except the case identified in #294, the captain train connector works perfectly
I let you build?


frankrousseau:
Great thx!

PR #296, opened 2016-06-01
Update free ISP operation labels

ZeHiro:
Free has updated their label. Previous label is kept for backward compatibility


frankrousseau:
Thx @ZeHiro

PR #301, opened 2016-06-04
Fix doctolib konnector

ZeHiro:


server/konnectors/doctolib.js
```diff
@@ -176,23 +176,23 @@ function saveEvents(requiredFields, entries, data, next) {
...
         // Create the event.
         connector.logger.info('Creating event');
         Event.create(icalEvent, (err) => {
           if (err) {
-            next(err);
+            callback(err);
           }
           entries.nbCreations++;
-          next();
+          callback();
```
frankrousseau: Be careful, her the callback is called twice if an error occurs.

ZeHiro: @frankrousseau 
Shall I return calling callback(err)?

frankrousseau: Yep

frankrousseau:
Thx @ZeHiro

PR #305, opened 2016-06-08
create APRR bills fetch

SomeAverageDev:


server/konnectors/aprr.js
```diff
@@ -0,0 +1,202 @@
...
+	headers: {
+        'Referer': 'https://espaceclient.aprr.fr/aprr/Pages/connexion.aspx',
+     },
+  };
+
+  connector.logger.info('Logging in on APRR website...');
+
+  request(options, (err, res, body) => {
+    if (err) return next(err);
+
```
ZeHiro: can you also detect when the credentials are wrong, and call `return next(err)`

server/konnectors/aprr.js
```diff
@@ -0,0 +1,202 @@
...
+  });
+};
+
+function parsePage(requiredFields, bills, data, next) {
+  bills.fetched = [];
+  moment.locale('fr');
+  const $ = cheerio.load(data.html);
+  var billFoundNumber = 0;
+
+	$('table[class=tbl_factures] tr').each(function (i, tr) {
```
ZeHiro: Did you run the linter? all this par seems to be unlinted.

server/konnectors/aprr.js
```diff
@@ -0,0 +1,202 @@
...
+
+		if (billUrl && billReference && billDate && billAmount) {
+			try {
+				billAmount = parseFloat(((billAmount.match(/(\d+,\d+)/))[0]).replace(',', '.'));
+
+				var month = parseInt((billUrl.trim().match(/processus=facture_(\d+)_\d+/))[1])-1;
+				billDate = billDate.split(' ');
+
+				var bill = {
+				  date: moment([billDate[1], month, 28]),
```
ZeHiro: Does this mean the bill is isued every 28th of the month?

server/konnectors/aprr.js
```diff
@@ -0,0 +1,202 @@
...
+				  vendor: 'APRR'
+				};
+
+				// saving bill
+				bills.fetched.push(bill);
+				billFoundNumber++;
+
+			}
+			catch (e) {
+				console.log(e);
```
ZeHiro: Can you delete the console statement? And call `return next(e)`

server/konnectors/aprr.js
```diff
@@ -0,0 +1,202 @@
...
+
+// Procedure to get hidden inputs
+function getHiddenInputs(requiredFields, bills, data, next) {
+  const url = 'https://espaceclient.aprr.fr/aprr/Pages/connexion.aspx';
+  const options = {
+    url,
+    method: 'GET',
+  };
+
+  connector.logger.info('Getting the hidden inputs...');
```
ZeHiro: Why don't you use the log variable you declared line 17?

ZeHiro:
@SomeAverageDev thank you very much for this new konnector. I have a few comments, can you address them? 
You also did not set the locales for the notification and the konnector. It expects 'konnector description aprr' and 'notification-konnectors-aprr'

server/konnectors/aprr.js
SomeAverageDev: hello @ZeHiro, there is no day for the bills, they are edited at the last day of the month, and 28 seems to be a good day for all months... I don't have a better idea.

server/konnectors/aprr.js
SomeAverageDev: I will try to check it !

server/konnectors/aprr.js
SomeAverageDev: I did not, how can I run it ?

server/konnectors/aprr.js
SomeAverageDev: will do

server/konnectors/aprr.js
ZeHiro: Npm run lint

server/konnectors/aprr.js
ZeHiro: Make sense

ZeHiro: Can you then add a comment please?

server/konnectors/aprr.js
ZeHiro: Thanks, The callback should be next('bad credentials')

server/konnectors/aprr.js
SomeAverageDev: will do
i did not check the integration within kresus, the parameters might need to be changed as the money transfer might appears at the begining of the following invoice month

      minDateDelta: 4,
      maxDateDelta: 20,

server/konnectors/aprr.js
SomeAverageDev: does this "return next(e)" will take me out of the each loop ? because I don't want to...

server/konnectors/aprr.js
SomeAverageDev: do you mean instead of connector.logger.info ? what is the best practice (I don't really need the log object) ?

server/konnectors/aprr.js
ZeHiro: I understood that. If you didn't test integration with Kresus, could you at least check the label of you operations (currently you set 'APRR' as the label, is it the one used by your bank. If several labels can be used, just pass an array instead of a string.

SomeAverageDev:
is there not an error in the SFR localization ?
in the key there is not an underscore:
const localizationKey = 'notification sfr box';

in the local file, there is an underscore
    'notification sfr_box': "%{smart_count} new invoice imported |||| %{smart_count} new invoices imported"

server/konnectors/aprr.js
ZeHiro: I wondered if that was what you wanted to do. Calling next will go to the next step il the list of functions you have established. So you shouldn't call next(e). Sorry

ZeHiro:
@SomeAverageDev : I think there is. I couldn't check this one (no SFR account)

server/konnectors/aprr.js
SomeAverageDev: hi,
i executed linting, and I have a question : which ES version should I be using ?
is cozy compatible with ES6 and for example templates (http://eslint.org/docs/rules/prefer-template) ?

server/konnectors/aprr.js
SomeAverageDev: the wensite returns HTTP 200 even if the login failed...
I can try to match a string to determine if the login worked but seems not so good...

do you have a smart idea to check this ?

SomeAverageDev:
hi
you can reject this pull, i made a new one (i had a problem with my git)

Issue #328, opened 2016-06-19
SNCF konnector broken


Chocobozzz:
They re made the orders page UI (snif). I'll make another PR to fix it.

PR #336, opened 2016-06-26
Fix sncf konnector (new html presentation)

Chocobozzz:
Some orders are in the old html form so I kept the old `getEvents` function. Once they will have completely migrated to the new html pages we will be able to remove it.

The new html order page is built with Angular which eases the parsing because we can directly fetch the JSON from the API.

I modified the linter to allow max line length to 120 because some JQuery selectors are very long.

Closes https://github.com/cozy-labs/konnectors/issues/328


frankrousseau:
Thank you @Chocobozzz

PR #349, opened 2016-07-07
Add application token for runabove

Chocobozzz:
I changed the ovh application tokens for a better application name/description.

I can't add tokens for Kimsufi fr/ca, Soyoustart fr/ca and ovh ca because I don't have an account on these services.

The OVH support said me that we can create an account on each service by creating an order and stop at the payment step.  Maybe I'm an idiot, but I tried and it never worked.

If there is no objection, I'll create an issue for each these services so anyone in the community that is client of one of them could create the associated application tokens.


frankrousseau:
Ok for the issues and thank you for the update!

PR #355, opened 2016-07-11
Fix podcast connector

frankrousseau:
Podcast connector seemed broken due to my recent changes. I made it work but I had to remove the promises, it made hard to 

Sometimes request can't find the feed. I wasn't able to understand why it happens.


babolivier:
> Sometimes request can't find the feed. I wasn't able to understand why it happens.

If you experienced this with a Trancendances podcast feed only, it's "normal", we have some unresolved issues on our RSS generation. So no problem from the connector.

PR #379, opened 2016-07-20
Fixing issue 351

SomeAverageDev:
Not the most elegant fix, but in order to make the konnector work, I will only enable the download of the lastest bill (the connector will download only one bill per fetch, the latest one...)
The user will need to fetch older bills by hand.

+linting APRR


ZeHiro:
Thanks @SomeAverageDev

PR #399, opened 2016-07-28
Fix SFR mobile (wrong date handling)

nicofrand:
If there is no year provided when using moment, moment will use the current year.
SFR does not provide the year for the last bill, but I guess they would if the bill's year was different from the current year.

This PR simplifies the code and fixes the previous issue (mix of Date/moment instances).


nicofrand:
@SomeAverageDev could you test it again please

SomeAverageDev:
All good this time, thanks !
All bills are identified and fetched with the right date.

nicofrand:
thanks @SomeAverageDev !!

PR #418, opened 2016-08-02
Fix free mobile

ZeHiro:


server/konnectors/freemobile.coffee
```diff
@@ -250,7 +250,7 @@ action=getFacture&format=dl&l="
     # - Import overall pdf with name YYYYMM_freemobile.pdf
 
     isMultiline = $('div[class="consommation"]').length > 1
-    $('div[class="factLigne hide "]').each ->
+    $('div[class="factLigne is-hidden "]').each ->
```
nicofrand: Why not "div.factLigne.is-hidden" ?

ZeHiro: Didn't try and didn't know the structure. I'll try tonight. This patch needs to be tested first

PR #419, opened 2016-08-02
Reset import Error on deletion of an account

### LINKED ISSUE #381: Removing a konnector's config should also remove the warning icon
**Steps to reproduce:**

0. Enter wrong id/password for digital ocean for example: a warning icon appears on its right in the menu and a "import error" message is displayed below the submit button.
0. Delete the configuration

**What happens:**

* The error message is still displayed
* The icon is still displayed.

**What I expect:**

The icon is removed as well as the import error message.

ZeHiro:
Fixes #381

This does not remove the import warning when deleting a specific configuration, but only when deleting the whole configuration (all the accounts attached to the konnector). It is not easy to have a "by configuration" management of the warning, as the importErrorMessage is managed on the konnector level, not an the account level.

### DIFF
--- a/client/app/views/konnector.coffee
+++ b/client/app/views/konnector.coffee
@@ -386,6 +386,8 @@ target="_blank">
                 @model.set 'lastAutoImport', null
                 @model.set 'accounts', [{}]
                 @model.set 'password', '{}'
+                @model.set 'importErrorMessage', null
+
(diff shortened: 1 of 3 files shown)


ZeHiro:
This is not enough. Something needs to be done on server side.

PR #429, opened 2016-08-03
Add vendor link on konnectors client. Fix #403

### LINKED ISSUE #403: In the konnector, having a link to the website connected
I was curious about the SNCF konnector and after some search in the source code I found that the konnector was designed to use http://www.voyages-sncf.com data and not http://www.sncf.com ones.

So I think, for the standard user, it will be better to click somewhere and see if it's the website he was thinking of.

There can be the same problem with localized website (OVH EU/US/...) or website with similar name (and tired peoples: https://www.showroomprive.com / https://vente-privee.com)
_ex:_ the user can't connect to his account
he follow the link and see he cannot connect to the website / he thought of an other website
he do not stay half an hour asking himself why he can not connect :)

ZeHiro:


client/app/views/konnector.coffee
```diff
@@ -372,10 +372,20 @@ target="_blank">
...
+            customViewElem = $ "<div class='customView'></div>"
             customViewElem.append translatedCustomView
             customViewElem.insertBefore @$('.fields')
 
+        if @model.has 'vendorLink'
+            link = @model.get 'vendorLink'
+            vendorLinkHtml = """
+<div class='vendorLink'>
+    <span>#{t 'vendorLink'}</span>
+    <a href="#{link}">#{link}</a>
```
nicofrand: Maybe the link should be escaped ? & converted to &amp; etc. ?

client/app/locales/fr.coffee
```diff
@@ -35,6 +35,8 @@ module.exports =
     'date format': 'DD/MM/YYYY [à] HH[h]mm'
     'add an account': "Ajouter un compte"
     'remove last account': "Supprimer le dernier compte"
+    'vendorLink': "Site d'où sont récupérées les données : "
+
```
nicofrand: There is a second blank line here I think…

PR #460, opened 2016-08-30
Fix #443 : remove unused 404 page

### LINKED ISSUE #443: 404 page unused
There appears to be a 404 page : https://github.com/cozy-labs/konnectors/blob/master/client/app/assets/404.html#L154 which is never called.

The CSS (`#goog` etc.) is obsolete too since the Google scripts were removed.

nicofrand:
Remove the 404 page: unused.
The reverse proxy must handle it and serve a [custom cozy error page](https://github.com/cozy/cozy-proxy/issues/297).


ZeHiro:
Thanks @nicofrand

PR #470, opened 2016-09-08
Update Ameli's konnector to match the last website version

Gara64:
This should fix the current Ameli's konnector state. I've successfully tested  with my account, but as usual, more tests are welcomed :)

### DIFF
--- a/server/konnectors/ameli.coffee
+++ b/server/konnectors/ameli.coffee
@@ -110,8 +110,9 @@ parsePage = (requiredFields, healthBills, data, next) ->
     startDate = $('#paiements_1dateDebut').attr('value')
     endDate = $('#paiements_1dateFin').attr('value')
 
-    billUrl = "https://assure.ameli.fr/PortailAS/paiements.do?actionEvt=" + \
-    "afficherPaiementsComplementaires&DateDebut="
+    baseUrl = "https://assure.ameli.fr/PortailAS/paiements.do?actionEvt="
+
+    billUrl = baseUrl + "afficherPaiementsComplementaires&DateDebut="
     billUrl += (startDate + "&DateFin=" + endDate)
     billUrl += "&Beneficiaire=tout_selectionner&afficherReleves=false&" + \
     "afficherIJ=false&afficherInva=false&afficherRentes=false&afficherRS=" + \
@@ -122,17 +123,18 @@ parsePage = (requiredFields, healthBills, data, next) ->
         strictSSL: false
         url: billUrl
 
-    # request the bill's url
     request billOptions, (err, res, body) ->
         if err
             log.error err
             return next 'request error'
         else
             $ = cheerio.load body
+            i = 0
+
+            # Each bloc represents a month that includes 0 to n reimbursement
             $('.blocParMois').each ->
-                pdfUrl = $($(this).find('.downReleve').get(0)).attr('href')
-                if pdfUrl?
-                    pdfUrl = "https://assure.ameli.fr" + pdfUrl
+
+                $('[id^=lignePaiement' + i++ + ']').each ->
 
                     amount = $($(this).find('.col-montant').get(0)).text()
                     amount = amount.replace(' €', '').replace(',','.')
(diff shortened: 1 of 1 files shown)


server/konnectors/ameli.coffee
```diff
@@ -98,56 +102,55 @@ parsePage = (requiredFields, healthBills, data, next) ->
...
-        bill =
-            amount: amount
-            type: 'health'
-            subtype: subtype
-            date: moment date, 'DD/MM/YYYY'
-            vendor: 'Ameli'
-            detailsUrl: detailsUrl
+    # request the bill's url
+    request billOptions, (err, res, body) ->
+        if err then next err
```
nicofrand: Please, can you log the error and return a locale key instead ?

```coffeescript
log.error err //where log is an instance of the logger of course
return next 'request error'
```

server/konnectors/ameli.coffee
```diff
@@ -80,9 +81,12 @@ logIn = (requiredFields, billInfos, data, next) ->
...
                     strictSSL: false
+                    headers:
+                        'Cookie': res.headers['set-cookie']
+                        'Referer': refererUrl
                     url: reimbursementUrl
 
-                # Last request to get the reimbursements
+                # Last request to get the reimbursements page
                 request reimbursementOptions, (err, res, body) ->
                     if err then next err
```
nicofrand: Please, can you log the error and return a locale key instead ?

```coffeescript
log.error err //where log is an instance of the logger of course
return next 'request error'
```

server/konnectors/ameli.coffee
```diff
@@ -98,56 +102,55 @@ parsePage = (requiredFields, healthBills, data, next) ->
...
-    options =
-        method: 'GET'
-        jar: true
-        strictSSL: false
-        url: detailsUrl
+                    healthBills.fetched.push bill if bill.amount?
 
-    # Request to get the pdf url
-    request options, (err, res, body) ->
-        if err? or res.statusCode isnt 200
```
nicofrand: Please, can you log the error and return a locale key instead ?

```coffeescript
log.error err //where log is an instance of the logger of course
return next 'request error'
```

nicofrand:
Thanks for the fix !!

Gara64:
@nicofrand Done!

nicofrand:
Thanks, I'll try to test it this evening too and then merge it.

nicofrand:
I can't build it and therefore can't test it:

```
Build entry point...

  rendered build/client/index.js

Entry point built.
Build locales...
/home/nicofrand/www/cozy/konnectors/client/app/locales/fr.coffee:16:6: error: reserved word 'import'
    'import already running': "L'import est déjà en cours."

```
I don't know why, I don't have the issue with the master branch.
Sorry for the delay. Would you mind trying `npm run build` to check if you have the issue too ?

nicofrand:
Sorry, but it did not seem to work. There was nothing in Files and in DataBrowser I have 1 entry with this value as pdfurl : http://localhost:12223/bill.pdf.

nicofrand:
I finally managed to make it work !
One small thing: the konnector seems to fetch the monthly resume PDF, not every reimbursement. The description should specify this, what do you think @Gara64 ?

Gara64:
Sorry for the delay, we're having a team week-end at Cozy :)
Actually, this is not good, I've realized that it only considers one reimbursement per month, no matter how many they are. 
So I'll fix this to deal with all reimbursements in the same month and download each pdf.

nicofrand:
I thought that was considering all reimbursements for a given month since it downloaded the summary of all reimbursements… It just depends whether you prefer one file per reimbursement or one per month… If per month, the description should be more accurate, that's all.

Thanks for the good work and have a great we with the cozy team !

Gara64:
@nicofrand I just pushed a commit that allows to fetch all the reimbursements and not just a monthly summary.
Took me some time, but the week-end was quite... intense :)

nicofrand:
Thanks a lot @Gara64 !

Issue #490, opened 2016-09-22
SNCF konnector makes the application crash


clochix:
Seen in a user's logs:
```
/usr/local/cozy/apps/konnectors/node_modules/konnectors/build/server/konnectors/sncf.js:170
    folders.forEach(function (folder) {
            ^
TypeError: folders.forEach is not a function
    at Request._callback (/usr/local/cozy/apps/konnectors/node_modules/konnectors/build/server/konnectors/sncf.js:170:13)
    at Request.self.callback (/usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/request/request.js:187:22)
    at emitTwo (events.js:87:13)
    at Request.emit (events.js:172:7)
    at Request.<anonymous> (/usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/request/request.js:1044:10)
    at emitOne (events.js:77:13)
    at Request.emit (events.js:169:7)
    at IncomingMessage.<anonymous> (/usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/request/request.js:965:12)
    at emitNone (events.js:72:20)
    at IncomingMessage.emit (events.js:166:7)
    at endReadableNT (_stream_readable.js:913:12)
    at nextTickCallbackWith2Args (node.js:442:9)
    at process._tickCallback (node.js:356:17)
```

Temporary workaround: added inside `getEvents`
```
if (!Array.isArray(folders)) return callback("No folders");
```

Side note: couldn't we add a `try-catch` block each time we call some method of a konnector, to prevent the whole application from crashing when there's an issue inside one konnector?

clochix:
Well, looks like folders is now an object with folder id as key.
I successfully imported the data adding the following patch:
```
    var folders = result.order.folders;
    if (folders && !Array.isArray(folders)) {
      var tmp = [];
      Object.keys(folders).forEach(function (ref) {
        tmp.push(folders[ref]);
      });
      folders = tmp;
  }
```

Ping @Chocobozzz

Chocobozzz:
Thanks @clochix. But I can't test because I don't have SNCF orders anymore.

Since it works for you, could you make a PR with this patch?

nicofrand:
```js
folders = Object.keys(folders).map(ref => folders[ref]);
```

Should be enough (I tried in my browser console only but seems to work without the temp variable).

ZeHiro:
@nicofrand can we close this issue?

[...]

PR #495, opened 2016-09-27
Fix crash of SNCF konnector

clochix:
refs #490


server/konnectors/voyages_sncf.js
```diff
@@ -200,6 +200,10 @@ function getEvents(orderInformations, events, callback) {
     }
 
     const folders = result.order.folders;
```
nicofrand: You need to use let instead of const here.

PR #505, opened 2016-10-03
WIP: Warn the user the password could not be parsed. If is moslty the case…

### LINKED ISSUE #503: Warn the user when the password is still encrypted.
When the DS and or the proxy are restarted, but not konnectors app, the password are encrypted (it is a bug in the DS/proxy) and konnectors tries to fetch data using the encrypted password as the real password. This leads to a fetch error. 
This should be prevented (see how Kresus manages that), and a notification shall be raised. 

Furthermore, it seems the "error" notification is deleted, which is an unexpected behaviour.

ZeHiro:
… when the data-system provides encrypted password. Fixes #503

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -1,6 +1,7 @@
 module.exports =
 
     'bad credentials': 'Bad credentials. Check the konnector fields and run the import again.'
+    'encrypted fields': 'The password(s) could not be decrypted. Please restart the application. If the problem persists, please contact the Konnectors maintener team' 
     'token not found': 'The token could not be retrieved'
     'key not found': 'Key not found'
     'request error': 'A request to the website failed, please see read the logs.'
--- a/client/app/locales/fr.coffee
+++ b/client/app/locales/fr.coffee
@@ -1,6 +1,7 @@
 module.exports =
 
     'bad credentials': "Mauvais identifiants. Vérifiez les champs que vous avez renseignés."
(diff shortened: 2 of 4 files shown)


client/app/locales/en.coffee
```diff
@@ -1,6 +1,7 @@
 module.exports =
 
     'bad credentials': 'Bad credentials. Check the konnector fields and run the import again.'
+    'encrypted fields': 'The password(s) could not be decrypted. Please restart the application. If the problem persists, please contact the Konnectors mainter team'
```
ZeHiro: OK

client/app/locales/fr.coffee
```diff
@@ -1,6 +1,7 @@
 module.exports =
 
     'bad credentials': "Mauvais identifiants. Vérifiez les champs que vous avez renseignés."
+    'encrypted fields': Le(s) mot(s) de passe n'a/ont pu être déchiffré(s). Veuillez redémarrer l'application. Si le problème persiste, veuillez contacter un mainteneur ed Konnector"
```
ZeHiro: OK

tests/password.coffee
```diff
@@ -62,6 +62,16 @@ describe 'Injecting/Removing encrypted fields', ->
                 for name, value of @fields
                     accountKeys.should.containEql name
 
+    describe 'Still encrypted fields', ->
+        it 'when I call injectEncryptedFields on a konnector with still enctypted fields' , ->
```
ZeHiro: OK

PR #507, opened 2016-10-04
Fix SFR mobile undefined urls (#497)

nicofrand:
I guess the webpage had an issue, this PR checks the url is incorrect nonetheless to avoid the connector to enter in a loop by being stuck in "importing".


SomeAverageDev:
works fine for me !

PR #510, opened 2016-10-13
Use local version of babel

ZeHiro:


nicofrand:
Did you test it ?

ZeHiro:
@nicofrand of course

nicofrand:
Thanks ! BTW you were ejected from the channel by the last netsplit

PR #513, opened 2016-10-13
Manage the case of unknown konnectors. Fixes #512

### LINKED ISSUE #512: Dev tools giving some undefined in error messages
Hi,

In [this function](https://github.com/cozy-labs/konnectors/blob/master/dev/connector.coffee#L23) I think it should be the connector name rather than the connector object that should be passed to the logged string.

Actually,

```bash
$ coffee dev_tools.coffee --status hghgzzeg
Moment Timezone 0.5.3 was already loaded with data from 2016c
error - Konnector dev tool | Can't find konnector undefined.
*******************************************************************************
```

which could be better without the "undefined" :)

ZeHiro:


nono:
@aenario can you look at this PR?

server/models/requests.coffee
```diff
@@ -1,16 +1,16 @@
-# See documentation on https://github.com/frankrousseau/americano-cozy/#requests
+# See documentation on https://github.com/frankrousseau/cozydb-cozy/#requests
```
aenario: The cozy-db url is https://github.com/cozy/cozy-db

ZeHiro: I saw it earlier, i'll correct. Thanks for the efficiency @nono a'd @aenario

Phyks:
I gave it a try, and it indeed solved the issues I had with the dev tools.

A remanining issue for now is that:

```
$ coffee dev_tools.coffee --status freemobile
Moment Timezone 0.5.3 was already loaded with data from 2016c
[2016-10-14 10:22:41:295] warn - Fields variable undefined, use curren one instead.
{ slug: 'freemobile',
  accounts: [],
  fields: undefined,
  lastSuccess: undefined,
  lastImport: undefined,
  isImporting: false,
  importInterval: 'none',
  importErrorMessage: null }
*******************************************************************************
```

it seems fields are not saved during the init.

ZeHiro:
@Phyks I have opened another issue for that: #515

PR #521, opened 2016-10-16
Add tests over konnector structure. Fixes #437. Fixes #475

### LINKED ISSUE #437: Forbid names with dots
The connector's name is used to display the connector's icon by creating a class with the same name.
Therefore, if the name is "foo.bar" the class is "foo.bar" too which then gives ".foo.bar" in CSS, which is invalid.

We should forbid the dots in connectors names and add a test to for it.

### LINKED ISSUE #475: Check that a connector's slug matches its filename
If the connector's slug (or the name which will be slugified) is different from the filename (without the extension), the connector is rejected and will not appear in the app view.

ZeHiro:
The failing tests are normal, there is a problem with the virgin mobile konnector. Should be fixed with #520

### DIFF
--- a/tests/konnector_structure.coffee
+++ b/tests/konnector_structure.coffee
@@ -1,0 +1,49 @@
+should = require 'should'
+fs = require 'fs'
+path = require 'path'
+slugify = require 'cozy-slug'
+
+konnectorsDirectory = path.resolve __dirname, '../server/konnectors/'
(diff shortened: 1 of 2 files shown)


nicofrand:
Maybe you could test that the filename matches the slug too ?
See https://github.com/cozy-labs/konnectors/issues/475

Issue #527, opened 2016-10-18
Add tests for missing translation keys for fields


ZeHiro:
if field is not password or login, some translation keys could be missing. Ex: loginUrl of OVH

PR #529, opened 2016-10-18
Add test to ensure the field of a konnector has a translation key. Fix…

ZeHiro:
…es #527.

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -42,6 +42,7 @@ module.exports =
     'vendorLink': "Website from which the data are imported: "
     'loginUrl': "Login URL"
     'token': "Token"
+    'refreshToken': "Refresh Token"
 
     'home headline': """
     With Konnectors you can retrieve many data and save them into your Cozy.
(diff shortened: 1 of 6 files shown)


client/app/locales/en.coffee
```diff
@@ -42,6 +42,8 @@ module.exports =
     'vendorLink': "Website from which the data are imported: "
     'loginUrl': "Login URL"
     'token': "Token"
+    'username': "User name"
```
nicofrand: There already is a login field, can't we use it instead of username ?

ZeHiro: Yes we can

ZeHiro:
I need to add the migration for the online_net konnector

client/app/locales/fr.coffee
```diff
@@ -42,6 +42,8 @@ module.exports =
     'vendorLink': "Site d'où sont récupérées les données : "
     'loginUrl': "URL d'authentification"
     'token': "Token"
+    'refreshToken': "Mettre à jour le token"
+
```
nicofrand: We should remove one of the blanklines, it would be easier to diff two language files. But that can be done after the merge

Issue #511, opened 2016-10-13
Labels are not translated


ZeHiro:
The labels of the bills should be in the user language (at the moment of the import).

ZeHiro:
We should also migrate existing labels.

nicofrand:
Not sure about that. Translating the label is kind of weird, if the user changes his language in the cozy platform he will have different labels for a same type of connector.

If the user finds it weird to have an english label, maybe the solution is not there:

* Maybe the user should be able to choose his own label for each connector (or even account) : probably the best solution
* Maybe the user should be able to migrate (in cozy, not konnectors) all the labels he want from one to another ("Bills" to "Factures" for example)

PR #531, opened 2016-10-19
Client-side architecture proposal

m4dz:
This PR introduces a technical doc about a revamp of the client-side application. Its main goal is to get feedback about technical choices, and let contributors make suggestions about our choices and their needs.

### DIFF
--- a/docs/client-side-architecture.md
+++ b/docs/client-side-architecture.md
@@ -1,0 +1,118 @@
+Cozy Konnectors client-side app architecture
+============================
+
+Current state
+-------------
+
+- Backbone app
+- Simple structure
+- Oignon architecture
+- Hard-coded templates structures (low flexibility)
+- Nested styles
+- Bugged realtime
+- Not a core to power konnectors, but not a full opionated plateform
+
+
+Should we keep it?
+------------------
+
+### Pros
+
+- it exists
+- it is used for dozen of konnectors
+- it doesn't need particular skills about the plateform to add a Konnector
+- it uses a well-known architecture (just a Backbone with some custom base views)
+
+
+### Cons
+
+- unmaintainable in long-term (markups in views, too deeply nested concerns…)
+- need to revamp all styles and a huge part of markups
+- doesn't offer enough flexibility to build views
+- adding features (like OAuth) will be difficult due to hard-coded behaviors
+- realtime isn't sufficiently sustainable (takes place at root instead of being dedicated to components it relies on)
+
+### Recommendation
+
+As the app isn't really complicated, and because we wanted to reconsider the way we display / manage / add features to konnectors, we recommand to trash the actual version of the client-side app, and start a fresh new one. The actual code is too deeply messy to be really sustainable in future, and doesn't let community easily add features.
+
+
+Architecture
+------------
+
+### Overview
+
+The konnector client-side app should provides the following features:
+
+- a dedicated core to power konnectors as autonomous entities
+- a set of useful components to build konnectors views (username, password, directory-select, tokens…)
+- a default view for basic konnectors (username / password / directory-target)
+- keep each konnector separated from the core
+- a minimalist API to interact with core features (e.g. to submit / validate form)
+- an easy way to override default behaviors (logics, views, styles…) for a given konnector
+
+
+### Technical recommendation
+
+We suggest to build the new version of Konnector client-side app using _Vue.js_ with _vue-router_ and _vue-resource_ for the app logics, and Webpack for the packaging tool along the Vue-loader webpack's loader.
+
+Vue.js is a frontend reactive framework useful to build good shaped interfaces and is built with design in mind. It doesn't need any complicated source logics (such as immutables sources, stores, etc) and instead keep focused on how to design components efficiently. Vue-router is its attached routing library, which makes URL based Ui really simple to develop.
+
+The Components approch of Vue.js, and its native filtering capabilities makes it a good candidate to build the new Konnectors Ui. It lets separate:
+
+- Main collection views (wizard, categories, configured konnectors…)
+- Ui components (input fields, secured fields, directory-selectors, interval-timer…)
+- Information subview (what does this konnector do?)
+- Configuration subview, which can be a dedicated one of the default built-in version
+
+A clever approach furbished by Vue.js ecosystem is the Vue-loader webpack's loader: it allows to  wrap in a single file all the needs to a dedicated component: template (markup), styles, and JS logics. This will allow community developers that need to create a custom view for their konnectors to wrap in one place all their behaviors, which will simplify contribution and maintenance, and will bring much more flexibility when we want to add features.
+
+
+### Core API
+
+The core will provide to each konnector component a set of methods (a minimal API) which allow it to interact with the functionnalities it needs. Those methods can be invoked using the `$emit` call (plus args) from the konnector component:
+
+#### `save`
+
+Send the data for the model to be saved.
+
+
+#### `validate`
+
+Ask the model to validate given data. Returns errors via `props`.
+
+
+#### `reset`
+
+Force the model to reset its data to default.
+
+
+#### `delete`
+
+Remove the model from the configured konnectors collection.
+
+
+### Components
+
+The core should provide a set of components ready to be used by konnectors:
+
+- input field
+- username field
+- password field
+- directory selector
+- interval timer selector
+
+
+### Default konnector view
+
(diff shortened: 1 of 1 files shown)


docs/client-side-architecture.md
```diff
@@ -0,0 +1,118 @@
...
+### Components
+
+The core should provide a set of components ready to be used by konnectors:
+
+- input field
+- username field
+- password field
+- directory selector
+- interval timer selector
+
```
nicofrand: Does that take multiple accounts into account too ?

docs/client-side-architecture.md
```diff
@@ -0,0 +1,118 @@
...
+### Default konnector view
+
+The default konnector view, which can be used by any konnector which doesn't need any custom form will provide the following fields:
+
+- username
+- password
+- directory
+- timer
+
```
nicofrand: Does that take multiple accounts into account too ?
What about apps that need tokens too ?

docs/client-side-architecture.md
m4dz: About multiple accounts, I think we can handle it the way data is stored in client app. Say, we have a _SNCF_ konnector, which allow only one account ; and a _Free_ konnector, which allow multiples. In the data source, the _SNCF_ key can be a plain object, configuring only one konnector, when _Free_ key contains an array, each item a plain object for an instance of this konnector.

This way, a view always configure only one instance of a given Konnector, but a Konnector can allow the user to configure multiple instances if needed.

docs/client-side-architecture.md
m4dz: IMO, tokens are an input field (maybe a password type? not sure at this time). If we develop a Konnector which need a token, we just have to create a dedicated Vue.js component that build a custom form, reusing the components offered by the core. It may be something like:

```vue
<template>
    <k-username/>
    <k-token />
    <k-interval/>
</template>
```

and it's done :)

nono:
If we start a new version of the client-side part of konnectors, I'd like that it will prepare the works for the cozy-stack-v2.

In this future environment, the client-side app and the konnectors will live in 2 repositories. So, it should be possible to add a new konnector by just adding code in the konnectors (as a server-side worker), without touching the client-side app. In my mind, the konnector will have a JSON description and this description should be enough for the client-side app to do its work.

When the client-side app is loaded, it fetches a list of all the available konnectors, and the configured accounts. Something like:

```json
{
  "konnectors": {
    "sncf": {
      "name": "Voyage SNCF",
      "icon": "/icons/sncf.svg",
      "description": "Importez vos factures et événements SNCF dans votre Cozy.",
      "categories": ["travel"],
      "inputs": {
        "username": {
          "type": "username",
          "label": "Identifiant"
        },
        "password": {
          "type": "password",
          "label": "Mot de passe"
        }
      },
      "accounts": [
        { "username": "Gustave", "password": "123456" }
      ]
    },
    "trainline": {
      // ...
    }
  }
}
```

The JSON description has to be more complex than in my example to handle i18n, the frequency for fetching data through the konnectors, and things like "Can the user configure several accounts of this type?".

docs/client-side-architecture.md
jsilvestre: text (username) and tokens are password field, right?

m4dz: @jsilvestre I'm not sure to understand: why username should be a password field?

m4dz: @ZeHiro custom / configurable selects seems legit, right 😃

m4dz:
I've just realized that you probably didn't see the mockups for the new version of the client-side app that @kelukelu has produced. The Marvel blueprint is here https://marvelapp.com/77d7jfg/screen/14185063.

@ZeHiro, about accessing to the differents accounts configured for a Konnector (aka, multi-accounts), I let @kelukelu explain what he have in mind 😃.

docs/client-side-architecture.md
nicofrand: I think the user should be able to define for each account/konnector his own tags, see https://github.com/cozy-labs/konnectors/issues/511#issuecomment-254748925.

The current state is that the developer chooses the tags (regardless of the language, etc.).

That should probably be studied too.

docs/client-side-architecture.md
m4dz: @ZeHiro I'm not sure to really understand what you've got in mind. Can you take a look at the @kelukelu mockup and tell me more?

docs/client-side-architecture.md
m4dz: @nicofrand we face the same problem in contacts about datapoint label (_home_, _work_, etc) which need to be customized by user, but translated for default choices (mentionned above). We'll try to find a clever solution for that, I hope we can reuse the same recipe here :)

nicofrand:
For the default tags just use English. The user will need to be able to
rename a tag somewhere in cozy, "that's all".

m4dz:
My initial thoughs was to provide a default view containing the most common fields (username / password / directory target / timer), and let konnector developer built its own view if more complex. But the suggestion @nono made before about configuring the konnector form from the backend endpoint is particularly sexy. I have some ideas to let Vue.js view to be dynamically setted from the JSON definition 😉.

I have a little doubt about translations, maybe they can be served by another endpoint, but it's a non-blocking point right now (IMHO).

I'll update the doc to reflect our discussion soon! Thanks a lot!

ZeHiro:
@m4dz the problem with translations is that if they are not present in the konnector, they need to be hardcoded in the client, which is not a good thing.

docs/client-side-architecture.md
jsilvestre: @m4dz poor wording from me: it's not about username field or password field, it's about type=text and type=password, whatever the label is, so your proposal looks fine from this point of view.

CPatchane: @doubleface It's expected to use ES6 for this new client app and remove the coffeescript usage.

m4dz:
I've just updated the doc to reflect your feedbacks, thanks again to all of you.

I've leave the translation apart, I don't think it's about the architecture right now, and I prefer we open another PR about it, especially when we'll have more to information about how we'll do it in the stack v2.

docs/client-side-architecture.md
```diff
@@ -0,0 +1,128 @@
...
+- a set of useful components to build konnectors views (username, password, directory-select, tokens…)
+- a default view for basic konnectors (username / password / directory-target)
+- keep each konnector separated from the core
+- a minimalist API to interact with core features (e.g. to submit / validate form)
+- an easy way to override default behaviors (logics, views, styles…) for a given konnector
+
+
+### Technical recommendation
+
+We suggest to build the new version of Konnector client-side app in ES6, using _Vue.js_ with _vue-router_ and _vue-resource_ for the app logics, and Webpack for the packaging tool along the Vue-loader webpack's loader.
```
misstick: Could you specify library used for these 4 cases; I explain myself:
 - missing with ES6 (implicit but not specified),
 - _Vue.js_: `vue.js_` for components,
 - _vue-router_: `vue-router.js_` for routing,
 - _vue-resource_: `_vue-resource.js_` for data managing.

(or something approaching)

Maybe adding framework links?

nicofrand: Btw, Phyks was asking on IRC if that made sense to develop konnectors using Vue.js whereas other cozy projects (email at least IIRC) are using React ?

m4dz: @misstick yep, I'll add the links to the different resources into the doc

@nono nothing more to add, just perfect, thanks!

Phyks: @nono ok, thanks for the feedback on this!

m4dz:
@nono can we consider the doc as OK and merge this PR?

nono:
@m4dz done

ZeHiro:
@nono, @m4dz should we also have an interface doc to specify the structure of a konnector in the new application?

m4dz:
@ZeHiro in a first step, Konnectors will be untouched, except to specify a category that they belongs to. I'll have to write a doc about it before the end of the week 😄.

At second time, when the stack v2 will be ready to serve konnectors, we'll produce some docs with @nono to specify the konnectors architecture and changes, and the API a konnector must be compatible with.

m4dz:
@nicofrand, @ZeHiro I'm a little bit confused: @nono merge this PR in oct 26th, into the `master` branch, and I can't find any trace of it into the history. You're the last to have pushed commits into `master` since. Have you seen something unusual? (I can of course re-open the PR, but I'm curious to understand :))

PR #462, opened 2016-09-06
Fix bad credentials error not displayed

nicofrand:
Fix #267 

- [x] ./googlecontacts.coffee
- [x] ./sncf.js
- [x] ./isen.coffee
- [ ] ~~./birthdays.js~~ not needed
- [ ] ~~./linkedin.js~~ won't fix, will fix when rewriting the connector
- [x] ./ovh_eu.coffee
- [x] ./ovh_ca.coffee
- [x] ./soyoustart_eu.coffee
- [x] ./soyoustart_ca.coffee
- [x] ./kimsufi_eu.coffee
- [x] ./kimsufi_ca.coffee
- [x] ./ical_feed.js
- [x] ./github.coffee
- [x] ./sfr_box.js
- [x] ./sfr_mobile.js
- [ ] ~~./nest.coffee~~ the API does not seem to allow to handle errors
- [x] ./online_net.coffee
- [x] ./runabove.coffee
- [x] ./rescuetime.coffee
- [x] ./digitalocean.coffee
- [x] ./bouyguesbox.coffee
- [x] ./bouyguestelecom.coffee
- [x] ./orange.coffee
- [x] ./sosh.coffee

PR #534, opened 2016-10-21
Fix ical feed connector after #462

ZeHiro:


server/konnectors/ical_feed.js
```diff
@@ -109,9 +109,12 @@ function saveEvents(requiredFields, entries, data, next) {
       }
     });
   }, (err) => {
+    if (err) {
```
nicofrand: Why would the error callback be called without an error ?
Is that normal ?

ZeHiro: Because the cb is called also when async.eachSeries has finished the table

PR #549, opened 2016-10-27
Add `vuejs` routing 

misstick:
- [x] add routes rules,
- [x] add vuejs components,
- [x] add CSS styles,
- [x] add translation,
- [x] add basic integration test on routing.


misstick:
@m4dz : Transifex is not working into jade template; could you explain me basic workflow plz?

client/app/components/category_list.vue
```diff
@@ -0,0 +1,5 @@
+<template lang="pug">
+    main(role="main")
```
m4dz: The `role=main` attribute is useless here as the `main` tag have this implicit semantic.

client/app/index.js
```diff
@@ -4,24 +4,36 @@ import VuePolyglot from './plugins/vue-polyglot'
...
+  },
+  {
+    path: '/discovery',
+    component: DiscoveryList
+  },
+  {
+    path: '/connected',
+    component: ConnectedList
+  }
+];
```
m4dz: If we stuck to [StandardJS](http://standardjs.com/) (upcoming in dev environment), no semicolon needed.

client/app/index.js
```diff
@@ -4,24 +4,36 @@ import VuePolyglot from './plugins/vue-polyglot'
...
-    new Vue({
-        router,
-        render: h => h(app)
-    }).$mount('[role=application]')
-})
+  new Vue({
+    router,
+    render: h => h(app)
+  }).$mount('[role=application]')
+});
```
m4dz: Same about semicolon :)

client/app/styles/base.styl
```diff
@@ -0,0 +1,91 @@
+@font-face
```
m4dz: `font-face` declaration isn't required as the page already load the global `font.css` file that contains all fonts declarations. (see #559)

client/app/styles/base.styl
```diff
@@ -0,0 +1,91 @@
...
+    font-family: 'SourceSansPro'
+    src: url('/fonts/source-sans-pro.ttf')
+    color: #333
+
```
m4dz: Tip: declaring default color in `@font-face` declaration isn't allowed

client/app/styles/base.styl
```diff
@@ -0,0 +1,91 @@
...
+    flex: 1
+
+[role="application"] main
+    flex: 1
+    padding: 2.5em 3em
+
+
+
+[role="application"] h1
+    font-size: 1.5em
```
m4dz: Tip: as styles files are preprocessed by _Stylus_, don't hesitate to nest selectors, it'll be more readable / maintainable

client/test/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+describe('(beta) test', () => {
```
m4dz: Could we commit an empty test file?

package.json
```diff
@@ -59,12 +59,20 @@
     "pug-cli": "1.0.0-alpha6",
     "should": "11.1.0",
     "sinon": "1.17.5",
-    "webpack": "1.13.2"
+    "webpack": "1.13.2",
+    "jade": "1.11.0",
```
m4dz: Rendering engine was migrated to pug in #555, could we please remove the legacy `jade` dep?

m4dz:
Thanks @misstick! I've a little doubt about base.styl, that declares much styles that are related to the `app.vue` component, and so should be in this component instead of a dedicated styles.

As a reminder, styles only files should only contains sets of rules (@extend) or default browser styles overrides, and then we could comsume them in components `*.vue` files directly.

Also, I recommend you take a look at https://github.com/cozy/cozy-ui/blob/master/src/stylus/cozy-ui/app/layouts.styl#L14 a it contains a ruleset for a layout that is near this one you want. You should probably create a new ruleset for the `2panes` layout, that we can externalize inside _cozy-ui_ later :)

client/app/app.vue
```diff
@@ -1,13 +1,77 @@
 <template lang="pug">
     div(role="application")
-        ul
-            li: router-link(to="/foo") foo
-            li: router-link(to="/bar") bar
+        aside(role="navigation")
```
m4dz: I think the `[role=navigation]` attribute should be on the `ul` above instead of the `aside` element (mainly because the `aside` contains a `h4` which isn't a navigational item)

client/app/app.vue
```diff
@@ -1,13 +1,77 @@
...
+            margin: 0 0 1em
+
+        h4
+            font-weight: normal
+            font-size: 1.5em
+            padding: 1em 1.5em
+            margin: 0
+
+
+      [role="navigation"]
```
m4dz: Maybe we can simply rely on the `aside` tag (or eventually be more specific with a `[role=application] > aside` and reduce the nesting of one level, it'll reduce the specificity :)

client/app/components/category_list.vue
```diff
@@ -0,0 +1,6 @@
+<template lang="pug">
+    main
+        div(role='contentinfo')
```
m4dz: The `[role=contentinfo]` could be put on the `main` tag, this div isn't usefull (except if it's here for layouts concerns)

client/app/components/connected_list.vue
```diff
@@ -0,0 +1,6 @@
+<template lang="pug">
+    main
+        div(role='contentinfo')
```
m4dz: Same about `[role=contentinfo]`

client/app/components/discovery_list.vue
```diff
@@ -0,0 +1,6 @@
+<template lang="pug">
+    main
+        div(role='contentinfo')
```
m4dz: Same about `[role=contentinfo]`

client/app/styles/base/_normalize.styl
```diff
@@ -0,0 +1,9 @@
+html
+body
+    margin: 0
+    padding: 0
```
m4dz: I'm not sure those are really needed as we already embed `normalize.css` which already sets `margin` for `body`

package.json
```diff
@@ -42,6 +42,7 @@
   "devDependencies": {
     "babel-cli": "6.11.4",
     "babel-preset-es2015": "6.6.0",
+    "babel-register": "^6.18.0",
```
m4dz: We should remove the `^` to set the version explicitely (instead of using a range of versions)

m4dz:
Thanks 🎉 !

PR #551, opened 2016-10-27
Fix env dev on clean clone

m4dz:
Fixes unwanted errors when we start a watch task fromn a clean environment (clean clone and empty DS)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "test": "sh scripts/lint.sh && sh scripts/tests.sh",
     "test:server": "sh scripts/tests.sh",
     "lint": "sh scripts/lint.sh",
+    "prewatch:client": "cd client && npm install",
     "start": "node build/server.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "test": "sh scripts/lint.sh && sh scripts/tests.sh",
     "test:server": "sh scripts/tests.sh",
     "lint": "sh scripts/lint.sh",
+    "prewatch:client": "cd client && npm install",
```
CPatchane: If I understand correctly, ```npm install``` will be run for each ```watch``` script? Is it necessary? I think it's good for the build script but maybe it's too much for the dev.

m4dz: It's our standard process, to ensure client deps are available. When client apps will become pure client-side apps and we avoid this messy double project repos, all of this shit will be dropped 😉

PR #555, opened 2016-10-28
Upgrade and improve tooling

m4dz:
This PR add config and upgrade template engine

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -25,9 +25,9 @@ echo "Previous client cleaned."
 
 echo "Build entry point..."
 mkdir -p ./build/server/views
-jade --client --no-debug --hierarchy --out ./build/server ./server
-echo "var jade = require('jade/runtime');module.exports=" | \
-    cat - ./build/server/views/index.js > ./build/server/views/index.js.tmp
+pug --client --no-debug --out ./build/server ./server
+echo "; module.exports = template" | \
+    cat ./build/server/views/index.js - > ./build/server/views/index.js.tmp
 mv ./build/server/views/index.js.tmp ./build/server/views/index.js
 echo "Entry point built."
 
--- a/server/views/index.pug
+++ b/server/views/index.pug
@@ -5,15 +5,15 @@ html(lang=locale)
   meta(name="viewport", content="width=device-width, initial-scale=1")
 
   title Cozy - Konnectors
-  
+
   link(rel="stylesheet" href="/fonts/fonts.css")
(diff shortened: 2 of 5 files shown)


server/views/index.pug
```diff
@@ -5,15 +5,14 @@ html(lang=locale)
...
 
   if imports
     script.
       window.initKonnectors = !{JSON.stringify(imports.konnectors)}
       window.initFolders = !{JSON.stringify(imports.folders)}
-  script(src="app#{hash}.js" defer=true)
+  script(src=`app${hash}.js` defer=true)
 
   div(role="application")
-
```
CPatchane: I think that we should keep the final empty line here like in all files.

m4dz:
Done, thanks!

CPatchane:
Thanks @m4dz!

scripts/build.sh
```diff
@@ -25,9 +25,9 @@ echo "Previous client cleaned."
 
 echo "Build entry point..."
 mkdir -p ./build/server/views
-jade --client --no-debug --hierarchy --out ./build/server ./server
-echo "var jade = require('jade/runtime');module.exports=" | \
-    cat - ./build/server/views/index.js > ./build/server/views/index.js.tmp
+pug --client --no-debug --out ./build/server ./server
```
ZeHiro: you should use the local version of `pug` -> `./node_modules/.bin/pug`

ZeHiro: same for babel/coffee etc

m4dz: Yep, that's the case, the build script export the `./node_modules/.bin` in front of `$PATH` ([here](https://github.com/cozy-labs/konnectors/pull/555/files/e1238dd339973ebd7fa54650fb9c41aebf0f9f9e#diff-d99b26f864526f6f8f79861d49deb922L3)), so all commands in the script are locals 😃.

ZeHiro: I'm tired :tired_face: I wonder how this would work with globally installed npm packages. Which one is prioritary? If I remember well, the global on is. Which can lead to tons of headaches :)

m4dz: > I'm tired 😫

😄 

The priority comes to the first available in the `$PATH`ENV variable, so with to top `export` instruction, your path become something like `./node_modules/.bin:/usr/bin:/bin`. So command is first seeking into the local `node_modules` binary dir, then fallback to `/usr/bin`(which contains generally links to globally installed node modules) and so on. No more headaches 😉

PR #558, opened 2016-10-28
Add client unit test environment

misstick:


package.json
```diff
@@ -52,17 +52,24 @@
...
     "express": "4.14.0",
     "minimist": "1.2.0",
-    "mocha": "3.0.2",
     "nock": "8.0.0",
     "npm-run-all": "3.1.1",
     "should": "11.1.0",
     "sinon": "1.17.5",
-    "webpack": "1.13.2"
+    "webpack": "1.13.2",
+    "jade": "1.11.0",
```
m4dz: We've migrated to `pug` in #555, can we remove the useless `jade` devDep?

package.json
```diff
@@ -52,17 +52,24 @@
...
+
+    "chai": "3.5.0",
+    "mocha": "3.0.2",
+    "jsdom": "9.6.0",
+    "jsdom-global": "2.1.0",
```
m4dz: Are `jsdom` really useful at this time? Instead of bloat deps, I suggest we try to keep them as focused as needed, and only add new (dev)Deps when we need them. Except if you consider `jsdom` as mandatory for unit testing :)

PR #546, opened 2016-10-26
Add webpack config and base Vue.js app

m4dz:
This PR introduces the base stack for developping the new Konnectors client-side app, according to #531. It:

- removes old useless `build` in development branch
- update buld script and add watcher tasks
- add a webpack config to build the front app, with Vue.js specifics
- add a _Hello World_ architecture to serves a default app, ready to dev

PR #560, opened 2016-11-01
Dev tools fixes

### LINKED ISSUE #547: Dev tools
Hi,

Just to sum up and centralize somewhere the status of dev tools and the documentation.
- [ ] [In the wiki](https://github.com/cozy-labs/konnectors/wiki/Use-the-dev-tools-to-make-the-development-easier) we should not use `freemobile` connector as it imports `PhoneBill` which is hardcoded in the module and does not allow the dev tools to show them. We should either use another konnector or fix `freemobile`.
- [x] Is there a way to reset the fields of a connector? (that is delete the accounts from the dev tools) `--reset` only purges models data I think.
- [x] Is it possible to `--init` multiple times (or is it conflict-prone?)?
- [x] In `coffee dev_tools.coffee --list`, I am not sure why there is an empty line, and no `vendor` listed.
- [x] There seems to be a problem with URL (or I did not find how to escape them):

```
$ coffee dev_tools.coffee --change weboob --fields weboobURL:http://localhost:8080/
info - Konnector dev tool | Changing fields for weboob with weboobURL:http://localhost:8080/.
Moment Timezone 0.5.3 was already loaded with data from 2016c
info - Konnector dev tool | Fields were successfully changed for weboob
Note: only fields described in the connector file are stored.

$ coffee dev_tools.coffee --status weboob
Moment Timezone 0.5.3 was already loaded with data from 2016c
[2016-10-26 14:45:31:296] warn - Fields variable undefined, use current one instead.
{ slug: 'weboob',
  accounts: [ { weboobURL: 'http' } ],
  fields: { weboobURL: 'text', JSONModulesDescription: 'text' },
  lastSuccess: undefined,
  lastImport: undefined,
  isImporting: false,
  importInterval: 'none',
  importErrorMessage: null }
*******************************************************************************
```

### LINKED ISSUE #553: Support konnectors to be a folder package
Hi,

It would be useful to be able to define a konnector as a folder with an `index.js` file exporting the right things.

Typical use case is my [Weboob PR](https://github.com/cozy-labs/konnectors/pull/548). This connector will be quite large, especially since I have to write translators between every Cozy models and Weboob types. It would be much more readable if I could do something like:

```
| /server/konnectors/weboob
    | index.js -> exports the weboob Konnector
    | /converters
        | Bill.js -> Contains converters to Bills Cozy type
        | ...
        | index.js
```

Not sure how difficult it is to support them in the server core though :/

Phyks:
* Fix split on `:` in dev tools which was preventing from setting a
field to a URL value. See
https://github.com/cozy-labs/konnectors/issues/547#issue-185478387.

* Use JSON description to set fields through the dev tools, which is a
more robust and reliable way to describe the fields values, without
parsing errors due to commas or colons.

* Fix `konnector_hash` and dev tools to be able to use ES6 konnectors as
well as JS folder import as a package.

* Introduce a new script to build server part only, which can be called
from `npm run build:server`.

* Convert dev tools code to ES6, using
[`decaffeinate`](https://github.com/decaffeinate/decaffeinate).

* Fix the dev tools to look for modules in `build` directory rather than
`server` directory directly. This makes it compatible with
ES6/Coffee/JS, at the expense of having to run a `npm run build:server`
before.

* Add a `npm run dev_tools` endpoint to ease launch of dev tools.

* `vendor` is not a Konnector model attribute and then should not be in
the ouput of `--list` command in the dev tools.

Fixes #547.
Fixes #553.


Phyks:
_Note_: I am not with my PC, and only have a bad internet connection at the moment, so I could not check that I did not introduce bugs in the rebase operation :/

Phyks:
Actually, it should be ok, I could test it with the cozy vm, even though I'd like another input on this :)

Purge operation is really long (more than 2 minutes with current konnectors), but I have no idea what takes so long :/

Phyks:
Another thing which might be interesting is a `npm run` command to deploy the local dev version of Konnectors to `cozy-dev`. I can put it in this PR as well, if you are ok with it.

dev/connector.js
```diff
@@ -0,0 +1,221 @@
...
+                return konnector.destroy(callback);
+            } else {
+                konnectors.map(konnector => konnector.destroy(callback));
+            }
+        });
+    }
+};
+
+
+function __guard__(value, transform) {
```
m4dz: Is this one really useful or can we write its call in a more smart way to remove it?

scripts/build_server.sh
```diff
@@ -0,0 +1,10 @@
+echo "Build server files..."
+echo "Building CoffeeScript..."
+./node_modules/.bin/coffee -cb --output build/server server
```
m4dz: Tip: I recommend your script override the `$PATH` env variable (see https://github.com/cozy-labs/konnectors/pull/546/commits/c066b61363976a5b283f110e0a3136ffd30a3619#diff-d99b26f864526f6f8f79861d49deb922R3), so you command call can simply be `coffee -cb ...` (without the local path)

scripts/build.sh
```diff
@@ -4,12 +4,7 @@ echo "Clean previous server"
 rm -rf build/server && mkdir -p build/server
 echo "Previous server cleaned."
 
-echo "Build server files..."
-./node_modules/.bin/coffee -cb --output build/server server
-./node_modules/.bin/coffee -cb --output build/ server.coffee
-./node_modules/.bin/babel ./server/konnectors -d build/server/konnectors
-./node_modules/.bin/babel ./server/lib -d build/server/lib
-echo "Server built."
+./scripts/build_server.sh
```
m4dz: I ask myself about the opportunity to hardcode paths. Shouldn't we use a `source $(dirname $0)/build_server.sh` here instead?

scripts/build_server.sh
```diff
@@ -0,0 +1,10 @@
+echo "Build server files..."
```
m4dz: Maybe we should prefix the file with the correct hashbang syntax `#!/usr/bin/env bash`?

server/controllers/konnectors.coffee
```diff
@@ -23,10 +23,17 @@ module.exports =
                     konnector.injectEncryptedFields()
 
                 # Add customView field
-                    konnectorModule = require "../konnectors/#{konnector.slug}"
-
-                    if konnectorModule.customView?
-                        konnector.customView = konnectorModule.customView
+                konnectorModule = require(
+                    path.join(
+                        "../konnectors/",
```
m4dz: If we rely on `path.join`, then a proper way to write it must be `path.join '..', 'konnectors', konnector.slug`, nope?

server/lib/importer.coffee
```diff
@@ -20,7 +20,14 @@ module.exports = (konnector, callback) ->
     # check if the konnector is created and if its not already importing
     if konnector.accounts?.length > 0 and konnector.isImporting is false
         log.info "Run import for #{konnector.slug}."
-        model = require "../konnectors/#{konnector.slug}"
+        model = require(
+            path.join(
```
m4dz: Same as previously about `path.join` proper split :)

server/lib/konnector_hash.coffee
```diff
@@ -2,7 +2,7 @@ fs = require 'fs'
 path = require 'path'
 
 currentPath = path.dirname fs.realpathSync __filename
-modulesPath = path.join currentPath, '..', 'konnectors'
+modulesPath = path.join currentPath, '../konnectors'
```
m4dz: This one comes in opposition to my last comments. By curiosity, why do you prefer the use of a `/` syntax instead of using `path.join` with one arg by path chunk?

scripts/build_server.sh
Phyks: Indeed, good idea. Altough, I have some concerns about it:
* Using `export` seems to be overkill, as it will pass down the `$PATH` to subprocesses, while we only need it in the current script I think?
* I think we should append to `$PATH` rather than prepend, in order to avoid having a system executable being replaced silently. This however means that the system-wide binaries will be preferred.

scripts/build_server.sh
Phyks: Good catch, will be fixed in next commit.

dev/connector.js
Phyks: Rewrote it in next commit.

scripts/build.sh
Phyks: Updated.

server/controllers/konnectors.coffee
Phyks: Indeed

server/lib/importer.coffee
Phyks: Indeed

server/lib/konnector_hash.coffee
Phyks: Updated. This change was made previously, and at the time I was needing to put something between the `..` and the `konnectors` part, hence it was easier to just edit a single string, if I remember correctly. Should be reverted indeed.

Phyks:
@m4dz thanks a lot for the review! I can squash the commits if you want.

m4dz:
> Actually it seems to be easier and better to simply do a npm start and head to http://localhost:9358.

We've introduced in the `development` branch a `npm run watch` task that launch bith server and client app in watch mode for an easy debugging. Things'll change when we'll go to the stack v2, so I suppose it's enough for now. Does it sounds good to you?

scripts/build_server.sh
m4dz: As export is scoped in the script only, I'm not sure it's a such nuclear weapon, and it let's commands rely on `$PATH` as usual :)

I prefer prepending to `$PATH`, so local versions installed by _npm_ are looked in first place. It ensures you always use the right version (e.g.: if I've installed coffee-script at system level and never updated it, appending to `$PATH` will produces a build with an outdated version, which may implies some unwanted errors). Using local versions allow us to finely control dev environments, a-la-virtualenv ;)

Phyks:
> We've introduced in the development branch a npm run watch task that launch bith server and client app in watch mode for an easy debugging. Things'll change when we'll go to the stack v2, so I suppose it's enough for now. Does it sounds good to you?

Perfect! No need for extra dev as it works at the moment and v2 is coming soon :p

scripts/build.sh
```diff
@@ -1,27 +1,28 @@
 #!/usr/bin/env bash
+PATH="$PATH:./node_modules/.bin"
```
m4dz: `./node_modules/.bin` should be prepend to $PATH to avoid system installed versions to take priority. It preserves to call wrong versions (especially with compilers like `jade` and `coffee` an introduced some bugs complicated to trace.

scripts/build.sh
```diff
@@ -1,27 +1,28 @@
...
 
 echo "Clean previous server"
 rm -rf build/server && mkdir -p build/server
 echo "Previous server cleaned."
 
-./scripts/build_server.sh
+source ./`dirname $0`/build_server.sh
```
m4dz: I think the right syntax avoid the `./`, nope? `dirname $0` let the script properly get the current path, though

m4dz:
Can you get the `dev_tools.js` script working? The es6 modules syntax causes it crashes locally. Have you some tips to help me to run it?

Phyks:
Oh, I run it through `npm` which handles everything: `npm run dev_tools -- [ARGS]`. It passes it to `babel-node` under the hood.

I can change this behavior if you think it is blocking.

m4dz:
*gosh* I'm definitely too tired, sorry ^^'… Thanks for the tip!

Phyks:
Np :) Thanks for reviewing it!

m4dz:
Al'right! Works pretty well 👍. Just to fix `build.sh` script and that's ready to merge 🎉

Phyks:
Should be good now.

m4dz:
Woohooo \o/ ! Thanks!

Phyks:
Thanks! Wiki should be in sync with this latest version of the dev tools: https://github.com/cozy-labs/konnectors/wiki/Use-the-dev-tools-to-make-the-development-easier.

aenario:
Hi @m4dz @Phyks ,

I have an issue : you use `source` in the build script to call build_server, but you call it with `sh` in the npm scripts. On ubuntu, `sh` is not `bash` but `dash` and I get a `source: not found`.

Phyks:
I think replacing `source` by `.` should be enough. @aenario could you test it and I'll fix it in my PR?

PR #568, opened 2016-11-02
Use-cases and Categories Documentation

m4dz:
This PR adds documentation about _use-cases_ and _categories_ behaviors in the new new client-side app.

📌 _Note about use-cases_: a _use-case_ is a way to sort available connectors in a different manner than simple categories, to provides to the user a way to discover some connectors organized by themes. You can see a mockup of the [_discover_ screen](https://marvelapp.com/77d7jfg/screen/14185063) (which presents configured _use-cases_), and a [_use-case_ screen](https://marvelapp.com/77d7jfg/screen/14184978) (accessible from the _discover_ screen, or in the onbaording) too.

### DIFF
--- a/docs/client-side-connector-configuration.md
+++ b/docs/client-side-connector-configuration.md
@@ -1,0 +1,71 @@
+
+---
+
+# Cozy MyAccounts: Connector configuration
+
+## Color
+
+The account connector can define a simple color or a gradient color to be used when displayed (for example as background). Here, it's important to know that the property is not a simple color but a css property. It's why the account connector color property will be defined like an object.
+
+We keep a simple color definition in all cases, because if we have a gradient as background, we can also need a simple for others interface elements (borders, customized svg...).
+
+For a simple color account connector:
+
+```javascript
+// my_connector.js
+
+    ...
+    color: {
+        // the simple color hexadecimal definition
+        hex: '#A7B5C6',
+        // property used to display the connector background in the modal,
+        // could be different from the hexColor property
+        css: '#A7B5C6'
+    }
+    ...
+```
+
+For a 'complex' color account connector:
+
+```javascript
+// my_connector.js
+
+    ...
+    name: "My connector"
+    fields: ...
+    models: ...
+    ...
+
+    color: {
+        // a default simple color still available, eventually for other usages
+        hex: '#9E0017',
+        // css property for linear gradient
+        css: 'linear-gradient(90deg, #EF0001 0%, #9E0017 100%)'
+    }
+    ...
+```
+
+If a color property is not defined by the account connector, that will fallback to the default ```hex``` and ```css``` value which is ```#A7B5C6```.
+
+## Category
+
+An account connector can define a category to be listed in. This category is single because a connected can not be listed in many different categories. Here is the connector category definition:
+
+```javascript
+// my_connector.js
+
+    ...
+    name: "My connector"
+    fields: ...
+    models: ...
+    ...
+
+    category: 'health',
+    ...
+```
+
+__⚠️ Important notes:__
+
+The defined category must be authorized by the MyAccounts application in order to be listed in. You can see more about the authorized categories in the [MyAccounts server side configuration documentation](server-side-configuration.md).
+
+If the account connector define a category which is authorized, it will be used. Otherwise, (if not 'valid' or not defined as well) that will fallback to the default name, which is ```others```.
--- a/docs/client-side-connector-configuration.md
+++ b/docs/client-side-connector-configuration.md
@@ -45,11 +45,11 @@ For a 'complex' color account connector:
     ...
 ```
 
-If a color property is not defined by the account connector, that will fallback to the default ```hex``` and ```css``` value which is ```#A7B5C6```.
+If a color property is not defined by the account connector, that will fallback to the default `hex` and `css` value which is `#A7B5C6`.
 
 ## Category
 
-An account connector can define a category to be listed in. This category is single because a connected can not be listed in many different categories. Here is the connector category definition:
+An account connector can define a category to be listed in. This category is single because a connector can not be listed in many different categories. Here is the connector category definition:
 
 ```javascript
 // my_connector.js
@@ -68,4 +68,4 @@ __⚠️ Important notes:__
 
 The defined category must be authorized by the MyAccounts application in order to be listed in. You can see more about the authorized categories in the [MyAccounts server side configuration documentation](server-side-configuration.md).
 
-If the account connector define a category which is authorized, it will be used. Otherwise, (if not 'valid' or not defined as well) that will fallback to the default name, which is ```others```.
+If the account connector define a category which is authorized, it will be used. Otherwise, (if not 'valid' or not defined as well) that will fallback to the default name, which is `others`.
--- a/docs/client-side-usecases configuration.md
+++ b/docs/client-side-usecases configuration.md
@@ -1,0 +1,105 @@
+last edit: 2016-11-02 11:13
+author: m4dz <code@m4dz.net>
+
+---
+
+
+Cozy MyAccounts Client Use-case
+===============================
+
+Abstract
+--------
+
+This document describes the context of using _use-cases_ in MyAccounts app, how they are related to the app and the stack, and how to customize them in hosting context.
+
+
+What are use-cases?
+-------------------
+
+A **use-case** is a way to sort and filter konnectors by an arbitrary denominator (e.g. _all konnectors that concerns billing_). It allow a user to browse available konnectors in a different way than browsing by categories. I allow the user to discover some konnectors by selecting first a _use-case_ (aka a scenario) that cover usages, and see what konnectors can be suggested.
+
+_MyAccounts_ can provides many _use-cases_, that can be configured, which mean a hoster can offer a different way to discover available konnectors (aka present different _use-cases_).
+
+
+How to define a use-case?
+-------------------------
+
+All _use-cases_ are declared in a manifest, built-in the client app, at `client/app/contexts/<context>/index.json`. Its architecture is defined ine the _Architecture of the manifest_ section below.
+
+The `context` value is the one defines in which context _MyAccounts_ app run. There must be a `cozy` context, which is the default one available. Later, all partners can defines their own _use-cases_ by adding a new context, with a dedicated manifest and locales (see _Localization_ section below about translations).
+
+Adding a _use-case_ simply mean adding a new object into the _use-cases_ array, that provides the following keys (see the _Architecture_ section below for a complete list and informations):
+
+- `slug`: `<String>` the slugname for the use-case
+- `konnectors`: `<Array[Obj]>` an array of objects defining the konnectors
+- `figure`: `<String>` the picture file that illustrate the use-case
+- `color`: `<Color>` a color object that defines the color of the use-case
+- `default`: `<Bool>` a `true|false` value that defines this use-case as the default one
+- `important`: `<Bool>` should this `use-case` a recommended one?
+
+For its first release, the manifest will be built-in the client app, to get it ready to work out-of-the-box. With the stack v2, we should serve the manifest behind a dedicated endpoint, so we won't need to build a specific version of _MyAccounts_ each time we need to provides different _use-cases_.
+
+
+Default values
+--------------
+
+There's two kind of default values for a _use-case_:
+
+### the `default` key
+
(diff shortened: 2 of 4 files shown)


m4dz:
@gregorylegarec I mention you so you can take a look at this one too (coz' I can't assign you: we need to talk :P)

docs/client-side-connector-configuration.md
```diff
@@ -0,0 +1,71 @@
...
+    color: {
+        // a default simple color still available, eventually for other usages
+        hex: '#9E0017',
+        // css property for linear gradient
+        css: 'linear-gradient(90deg, #EF0001 0%, #9E0017 100%)'
+    }
+    ...
+```
+
+If a color property is not defined by the account connector, that will fallback to the default ```hex``` and ```css``` value which is ```#A7B5C6```.
```
m4dz: Tip: you can use simple backtick for inline `code` tag

docs/client-side-connector-configuration.md
```diff
@@ -0,0 +1,71 @@
+
```
m4dz: maybe we could add `author` and `date` metas (@nono, any though about this practice to add metas in YAML in front of docs mardown files?)

docs/client-side-connector-configuration.md
```diff
@@ -0,0 +1,71 @@
...
+        css: 'linear-gradient(90deg, #EF0001 0%, #9E0017 100%)'
+    }
+    ...
+```
+
+If a color property is not defined by the account connector, that will fallback to the default ```hex``` and ```css``` value which is ```#A7B5C6```.
+
+## Category
+
+An account connector can define a category to be listed in. This category is single because a connected can not be listed in many different categories. Here is the connector category definition:
```
m4dz: Typo: "because a **connector** can not be listed"

docs/server-side-configuration.md
```diff
@@ -0,0 +1,37 @@
...
+
+The MyAccounts categories names list won't be automatically computed from available connectors in order to keep it clear and relevant for all Cozy users with a growing number of connectors.
+
+Instead, they need to be configured on the served side, in a simple way, from a config file. This is a list of categories authorized to be used:
+
+```javascript
+{
+    authorizedCategories:
+        {
+            'insurance': 1,
```
m4dz: I'm not sure why we declare categories as object keys. I think an array of objects, containing at least a `slug` key would be easier to maintain and to enrich if needed.

By curiosity, what have you in mind when declaring `<categoryName> : 1`? (and so, what does `1` means here?)

docs/client-side-connector-configuration.md
CPatchane: Thanks for the tip, I didn't know it 👍

docs/server-side-configuration.md
CPatchane: @m4dz I just thought here about a list of authorized categories as an object with slug names as properties. This list is just used to lookup in, if a category is 'authorized' to be used in the client application or not. The `1` values are just to give a truthy value.

__Why an object?__ Actually for two main reasons, all related to the fact that this list will be used only to lookup in:

* Lookup in a list is more readable using an object than an array:

    ```coffee
    if not authorizedCategories[category]
        # do something

    # VS

    if authorisedCategories.indexOf(category) isnt -1
        # do something
    ```

    Even if this part will be solved with the ES7 `includes` function, there is another point.

* For performance reason. Use an object to lookup in thanks to property access is so much faster than lookup into an array using `indexOf`. You can test this performance with this [jsperf](https://jsperf.com/array-indexof-vs-object-key-lookup2). And the ES7 `includes` is not much faster for now ([jsperf](https://jsperf.com/lodash-includes-vs-indexof/2)).

Of course, it's my opinion and all of that is open to discussions.

docs/client-side-connector-configuration.md
CPatchane: You're right, I'll fix it, thanks!

docs/server-side-configuration.md
nono: I'd start with a just an array of slugs. It is what is easier to read & edit.

> Lookup in a list is more readable using an object than an array:

Not really. ES7 has `includes`. Coffescript gives you:

```coffee
if not category in authorizedCategories
    # do something
```

And the right way to do a lookup on an object is:

```coffee
if authorizedCategories.hasOwnProperty(category)
    # do something
```

Else, wait for the day someone will create a konnector with the [`watch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/watch) category :see_no_evil:

> For performance reason.

Well, if a konnector can be in hundeds or thousands of categories, an object will be faster than an array. But, with very few categories, it's not the case.

CPatchane: Great @nono! Thanks for these explanations. I completely forgot the object prototype properties here and to use hasOwnProperty, my bad :/. Finally, it's not readable as before.

Also agree with the coffescript usage here on server side. 

I just think about performance here because the category checking will be processed for each konnector each time the client will request it. Of course, it's not so much important for us here with few konnectors and few categories.

docs/client-side-connector-configuration.md
m4dz: Dully noted, thanks :)

gregorylegarec:
Typo : looks like `client-side-usecases configuration.md` lacks a dash between _usecases_ and _configuration_.

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+Abstract
+--------
+
+This document describes the context of using _use-cases_ in MyAccounts app, how they are related to the app and the stack, and how to customize them in hosting context.
+
+
+What are use-cases?
+-------------------
+
+A **use-case** is a way to sort and filter konnectors by an arbitrary denominator (e.g. _all konnectors that concerns billing_). It allow a user to browse available konnectors in a different way than browsing by categories. I allow the user to discover some konnectors by selecting first a _use-case_ (aka a scenario) that cover usages, and see what konnectors can be suggested.
```
CPatchane: Typos: I think that you mean "It allows" in the two next sentences here.

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+_MyAccounts_ can provides many _use-cases_, that can be configured, which mean a hoster can offer a different way to discover available konnectors (aka present different _use-cases_).
```
CPatchane: Typos: "can _provide_", "which _means that an hoster_ can offer"

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+How to define a use-case?
+-------------------------
+
+All _use-cases_ are declared in a manifest, built-in the client app, at `client/app/contexts/<context>/index.json`. Its architecture is defined ine the _Architecture of the manifest_ section below.
```
CPatchane: Typo: "is defined _in_"

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+The `context` value is the one defines in which context _MyAccounts_ app run. There must be a `cozy` context, which is the default one available. Later, all partners can defines their own _use-cases_ by adding a new context, with a dedicated manifest and locales (see _Localization_ section below about translations).
```
CPatchane: Typo: I think that you mean "is the one _that_ defines"

Typo: "all partners can _define_"

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+There's two kind of default values for a _use-case_:
+
+### the `default` key
+
+it defines if this _use-case_ is the default one. When we access the _use-cases_ screen (`/discover` URI), it offers a list of _use-cases_ screen, each one redirecting to a _use-case_ view (or the konnector if the _use-case_ only contains one).
+
+Sometimes, we need to directly open a _use-case_ screen without passing by the `/discover` view (this is the case when user access to _MyAccounts_ from the onboarding). In this case, this is the default _use-case_ which is displayed.
+
+⚠️ if the _use-cases_ array contains more than one _use-case_ with a default key set to `true`, then only the first one find in the array is considered as the default one.
```
CPatchane: Typo: "the first one _found_"

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+### The `incentive`
+
+When displaying a _use-case_ screen, a konnector can be highlighted first to incitate the user to first configure this one. Into the `konnectors` array, the _Konnector Object_ can define a `default` key at `true` to declares it as the _incentive_ one.
```
CPatchane: Typo: "to _declare_ it"

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+
+The `context` value is the one defines in which context _MyAccounts_ app run. There must be a `cozy` context, which is the default one available. Later, all partners can defines their own _use-cases_ by adding a new context, with a dedicated manifest and locales (see _Localization_ section below about translations).
+
+Adding a _use-case_ simply mean adding a new object into the _use-cases_ array, that provides the following keys (see the _Architecture_ section below for a complete list and informations):
+
+- `slug`: `<String>` the slugname for the use-case
+- `konnectors`: `<Array[Obj]>` an array of objects defining the konnectors
+- `figure`: `<String>` the picture file that illustrate the use-case
+- `color`: `<Color>` a color object that defines the color of the use-case
+- `default`: `<Bool>` a `true|false` value that defines this use-case as the default one
```
gregorylegarec: Are we ok that the default value for `default` property should be `false`, so we have only one use case containing this property (set to `true`) ?

docs/client-side-usecases configuration.md
gregorylegarec: It it the reference to `default` a typo ? Aren't we talking about `incentive` property ?

gregorylegarec: However : does `incentive` property refers to `important` defined in previous paragraph ?

docs/client-side-usecases configuration.md
gregorylegarec: I am not a big fan of the file name : `index.json`. I think it is too generic and means everything and nothing. Why not call it `manifest.json` ? It is more self-descriptive in my opinion. What yours ?

docs/client-side-usecases configuration.md
```diff
@@ -0,0 +1,105 @@
...
+    "color": {
+      "css": "the CSS value of the use-case color"
+    },
+    "default": true,
+    "important": true
+  }]
+}
+```
+
+The `important` key indicate if the _use-case_/_konnector_ is recommended or not.
```
gregorylegarec: Those last two property descriptions seem like a repetition of what has been previously told int the document. What do you think ?

docs/client-side-usecases configuration.md
m4dz: My opinion is that, with an `index.json` file, we can safely do a `const ctx = require('app/contexts/<mycontext>')` to load a context because the require will try to load an `index` file by default, and I find this way very smart. But I'm open to any other suggestion :)

docs/client-side-usecases configuration.md
m4dz: Yeap, totaly agree with that 👍

docs/client-side-usecases configuration.md
m4dz: Nope, my first though was to have an incentive property. But as we must have konnectors defined as objects in the `konnectors` key (to mark those _importants_), I finally think that the incentive one (if there's one) should be marked itself as `default` in the `konnectors` array. Does it sounds good?

The `important` property is here to mark a _use-case_ / a  _konnector_ with the `recommended` label. Maybe we shoudl change the property name?

docs/client-side-usecases configuration.md
m4dz: As described above, you can mak a **use-case** as `important` (aka _recommended_ see [_use-case_ screen](https://marvelapp.com/77d7jfg/screen/14185063)) and in a _use-case_, some **konnectors** can be marked as `important` too ([here in a category](https://marvelapp.com/77d7jfg/screen/14295255), but same things happened with _use-case_ content (poke @kelukelu to confirm this one))

docs/client-side-usecases configuration.md
gregorylegarec: You're right, OK for me.

docs/client-side-usecases configuration.md
gregorylegarec: So, ok for me !

PR #571, opened 2016-11-03
Add category and color properties to all konnectors

CPatchane:
Update all konnectors to add a category and a color, following the documentation in PR#568


ZeHiro:
@CPatchane are these 2 new parameters compulsory? Is so could you update the tests accordingly ?

CPatchane:
@ZeHiro Those are two new parameters, they are recommended (for customizing) but not compulsory. The server side will have fallbacks to handle the case when they are missing/invalid and will use default values instead.
These fallbacks are defined in the PR #570.

m4dz:
Thanks 🎉 !

PR #572, opened 2016-11-03
Proposition for generic redirection handling in connectors. 

jacquarg:
No multi-accounts, bad konnectors settings handling, bad UI. Try facebook_event.

What the principle: allow the konnector to get some token from a redirect url.

How it works:
1. A new konnector property: connectUrl , to the set the authentification endpoint URL. This activate a Connect button (link) in the konnector UI.
2. A new route : konnectors/:konnectorId/redirect , which store the data from the query-string of the url, and run the konnector.

To be done / discussed : 
multi-account <-- DONE
~~improve UX : keep only one action button, as in the other konnectors.~~


ZeHiro:
Nice @jacquarg can you see if this solution could be used also for the Google-Contacts connector? 
I think the multiple account won't be a problem (to be checked)

jacquarg:
Hi @ZeHiro,
Yes it will be usable with Google-Contacts connector, and many others OAuth2 API (LinkedIn, Uber, ...). It doesn't need to copy and past a token, but, as the Facebook events konnectors, those will probably need a redirection proxy (a simple static page like this: https://github.com/jacquarg/jacquarg.github.io/tree/master/proxy_redirect/oauth2 .

For the multi account point ; do you advice to desactivate it for this kind of _redirect konnectors_ ; or do you think it should be easy to achieve ?

Phyks:
Shouldn't the proxy be built-in as an endpoint on the server-side?

jacquarg:
Hi, 
I just have made some changes to allow multi-account with this kind of konnectors.

@Phyks , I don't understand your question. The role of the proxy is to offer a fixed URL for one konnector (and not one per user as with Cozy), to comply with the restrictions of many OAuth2 endpoints (such as Facebook, Google, ...), which wait for one url, defined during the registration process of the konnector (the app), in their respective developer consoles. 
Considering that, I don't understand well on which server you propose to put the proxy  : )

Phyks:
@jacquarg Sorry, I missed your message, and I misread the purpose of the proxy :/

Having a look at [https://github.com/jacquarg/konnectors/blob/6526a9cb0488a611d5c4603b66ec236f8258825c/server/konnectors/facebook_events.js#L49](this), my concern was that it meant a third-party server was having access to the `accessToken` if I understand correctly. I wanted this page to be served by a route in Konnectors, but that indeed is not doable as a user would have to register itself for an app token on Facebook API, if I understood correctly :/

jacquarg:
Thank's @aenario for approval !

@Phyks : yes, that is the weak point of the solution: most of OAuth2 endpoint will need a proxy. For the specific case of Facebook, as everything happen in client side (redirection with fragment, and not querystring), I am almost certain that the third party (github in that particular case) doesn't see any token. But, it's not obvious, and will not be systematic with each OAuth2 endpoints. We may notice that there is a redirection proxy clearly to the user.

Phyks:
@jacquarg It nevers sees it, unless the served source code is not the expected one. Github can typically serve a slightly different page on instead of https://github.com/jacquarg/jacquarg.github.io/blob/master/proxy_redirect/facebook_events/index.html, making an AJAX call to their server to intercept the token. I do not see much other solutions though, to keep it user friendly.

Maybe @cozy could serve this proxy page on their servers, behind a `*.cozy.io` domain?

@jacquarg One interesting point though would be to have it easily configurable, so that a user who is self-hosting and is ok to generate his own API tokens could use them?

poupotte:
@jacquarg : I tested this PR with branch [572-build](https://github.com/cozy-labs/konnectors/tree/572-build/server) .

When I try to connect my facebook konnector, I receive some errors :/ :

In my javascript console : 
```
PUT 
XHR 
https://zoe.cozycloud.cc/apps/konnectors/konnectors/da1c317cab404ac6250096207d3ac73c [HTTP/1.1 500 Internal Server Error 41 ms]
11:21:47,067 GET 
XHR 
https://zoe.cozycloud.cc/apps/konnectors/socket.io/ [HTTP/1.1 500 Internal Server Error 18 ms]
```

In my server logs, you can see that konnectors is restarted (but there isn't errors) :
```
[2016-11-14 10:18:39:60] info - americano | Configuration for production loaded.
konnectors server is listening on port 9358...
info - realtime-adapter | socket.io initialized
[2016-11-14 10:18:39:206] info - init | All konnectors successfully migrated
[2016-11-14 10:18:39:216] info - init | Successfully migrated all fields
info - konnectors | Import poller started.
[Mon, 14 Nov 2016 10:21:32 GMT] - GET / - 200 - 115.235 ms -  36955
[2016-11-14 10:21:46:652] info - americano | add plugin: cozydb
[2016-11-14 10:21:46:657] info - Cozy DB | bill - byDate request creation...
[2016-11-14 10:21:46:727] info - Cozy DB | succeeded
[2016-11-14 10:21:46:727] info - Cozy DB | bill - byVendor request creation...
[2016-11-14 10:21:46:759] info - Cozy DB | succeeded
[2016-11-14 10:21:46:759] info - Cozy DB | konnector - all request creation...
``` 
```

jacquarg:
Thanks @poupotte for testing !
I just test 572-build (but in development mode), and get stuck on a error to : after the redirections from facebook, konnectors crashes.

It comes from this line : https://github.com/cozy-labs/konnectors/blob/572-build/server/controllers/konnectors.coffee#L27
path is not declared. This were introduced with this commit : https://github.com/cozy-labs/konnectors/commit/0fb27a86a233a1178c2c02c191567f30dca0d373#diff-63fdd83b59ac40571ccb831f2fce935fR27 (ie #560 ).

How do I fix this ? in a new PR ?

poupotte:
New PR seems good :)

ZeHiro:
@jacquarg , @poupotte isn't this PR "conflicting" with what being done here : #580 or is it complementary (or they are completly different?)

jacquarg:
@poupotte, Github prefers this PR ...

@ZeHiro , if #580 is what I think, it will be complementary, probably with some conflicts.. : ) 
THe thing, is that we need the functionality of this PR ready in master soon, certainly some weeks before the #580 is ready.

Phyks:
@jacquarg just use another branch on your fork. Github limits to one PR per branch on the fork.

Anyways, as I introduced it, I can also write a PR to fix it. Let me know if you prefer it.

aenario:
@poupotte 572-build branch up to date with this PR. 
The crash was introduced in #560 (and not #540 as indicated in commit message).

The crash is **still on master** and will appears at next build if it is not from this (572) PR which now contains the fix.

jacquarg:
@aenario , thanks, I just fixed the error in the commit label : )

poupotte:
@aenario : it works with news commits (330cfda)

PR #574, opened 2016-11-03
Add MesInfos EDF Konnectors

jacquarg:
Need to add the EDF icon.


server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1358 @@
...
+    statementReason: String
+    period: String
+    cost: Number
+    costsByCategory: Object
+    valuesByCatergory: Object
+    similarHomes: Object
+    releve: [Object]
+    docTypeVersion: String
+
+Bill = cozydb.getModel 'Bill',
```
ZeHiro: Could you use the generic bill model?
The same could be done for the other models (to be thought of, as they seem very specific to EDF)

package.json
```diff
@@ -38,7 +38,9 @@
     "querystring": "0.2.0",
     "request": "2.74.0",
     "request-json": "0.6.1",
-    "ware": "1.3.0"
+    "ware": "1.3.0",
```
ZeHiro: The authorisations for the new docTypes are not declared.

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1358 @@
...
+                'ns:valeur': client.clientId
+            ,
+                'ns:cle': 4
+                'ns:valeur': billNumber
+            ,
+                'ns:cle': 6
+                'ns:valeur': 'Facture'
+            ]
+
+    edfRequestPost path, body, (err, result) ->
```
ZeHiro: Could you use save_data_and_files library for that?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1358 @@
...
+prepareEntries = (requiredFields, entries, data, next) ->
+    entries.homes = []
+    entries.consumptionStatements = []
+    entries.contracts = []
+    entries.bills = []
+    # entries.clients = []
+    # entries.paymentTerms = []
+    next()
+
+updateOrCreateDocs = (requiredFields, entries, data, next) ->
```
ZeHiro: can you use filter_existing and save_data_and_files libs?

nicofrand:
Hi, thanks for this PR which will be useful for a lot of people, no doubt.

However, I find it a bit a pity that coffeescript is used instead of ES2015.
Coffeescript is quite hard to read, debug, and has led us to several issues in the past.

The next version of cozy (as in cozy stack v2) will only use ES2015 (when using JS of course) for the client side, but since konnectors will be the only one with JS on the server side, it would make sense to remove any Coffeescript dependency.

I know that would probably be a lot of work, but would you be so kind to consider revising your PR and use ES6, please ?

Phyks:
About https://github.com/cozy-labs/konnectors/pull/574#issuecomment-258187936, http://decaffeinate-project.org/ might ease a lot the conversion of the files of this PR.

jacquarg:
Thank for your comments.
I try to move to ES6, decaffeinate works greate, but after I got 500 Lint errors ; so I put it in an other branch, and work on it later.

I just make some proposition on Bill model, to use the common one.
I will have a look on the other comments now.

server/konnectors/edf.coffee
jacquarg: Sadly no, because I have to use the specific token and request format for the request.

server/konnectors/edf.coffee
jacquarg: No, I can't use filter_existing because:
1. I need custom criteria on filtering, specific for each DocType
2. I need updateOrCreate, and save_data does only createIfNotExist

But I could put updateOrCreate in a lib, it may become a usefull tool for others ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1347 @@
...
+
+# Models
+Client = cozydb.getModel 'Client',
+    clientId: String
+    vendor: String
+    numeroAcc: String
+    address: Object
+    name: Object
+    email: String
+    telMobile: String
```
Phyks: Some fields here and below are named in French, others are named in English (typically `commercialContract` vers `telMobile` or `numeroDepannage`).

server/konnectors/edf.coffee
Phyks: Related to https://github.com/cozy-labs/konnectors/issues/575 (mentionning to keep track of this in the issue).

server/models/bill.coffee
```diff
@@ -13,6 +13,10 @@ module.exports = Bill = cozydb.getModel 'Bill',
     fileId: String
     content: String
     isRefund: Boolean
+    clientId: String # Client number in vendor CRM
+    number: String # Bill Id, in vendor CRM
```
Phyks: Maybe rename it as `vendorBillId` (seems clearer to me about what it should contain, `number` could be understood as `amount`). Also, what is it useful for?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1347 @@
...
+
+parser = new xml2js.Parser()
+builder = new xml2js.Builder headless: true
+
+# TODO :
+# Move updateOrCreate in a lib
+# multi account
+# multi contract
+
+# Models
```
Phyks: There are a lot of new models introduced here, and I fear they may be too specific? Not sure what is the standard Konnectors approach here. Typically, `Contract` is a kind of translation of a [`Subscription`](http://dev.weboob.org/api/capabilities/bill) model in Weboob, but this one is much more specific.

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1347 @@
...
+    modifIbanAutorisee: Boolean
+    dernierReglement: Object
+    periodiciteFacture: String
+    dateProchaineFacture: String
+    idPayeur: String
+    payeurDivergent: Boolean
+    paymentSchedules: [Object]
+    docTypeVersion: String
+
+Home = cozydb.getModel 'Home',
```
Phyks: Isn't the name too generic?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1347 @@
...
+    occupationType: String
+    constructionDate: String
+    isBBC: Boolean
+    surface: Number
+    occupantsCount: Number
+    principalHeatingSystemType: String
+    sanitoryHotWaterType: String
+    docTypeVersion: String
+
+ConsumptionStatement = cozydb.getModel 'ConsumptionStatement',
```
Phyks: Same statement as above, it can be compared with the [`Details`](http://dev.weboob.org/api/capabilities/bill) model in Weboob, but once again is much more specific.

server/models/bill.coffee
```diff
@@ -13,6 +13,10 @@ module.exports = Bill = cozydb.getModel 'Bill',
     fileId: String
     content: String
     isRefund: Boolean
+    clientId: String # Client number in vendor CRM
```
Phyks: Should not it be linked to the contract rather than to the client id?

server/konnectors/edf.coffee
Phyks: Also, could you put some comments describing what each field should contain?

jacquarg: Yes there is a lot of models, very specific : the whole list [here](http://mesinfos.fing.org/cartographies/datapilote/).
But that is just a great news ! There will be a lot of different and usefull data in the Cozy, so Cozy's users will access to a lot of very interesting and usefull apps reusing them ; ) 

About the specificity to this specific vendor : yes, today it's really specific, and sometimes not well named or structured. We could follow some standard when we found them, but we shouldn't loose data. The Cozy arch allows us to have as much specific field has we want on each documents, and we have to use this feature. But that wy I put a docTypeVersion on each document ; to help developpers track document, and the meaning of their fields, even after models updates, and variations on model's fields.

About the contract model: the name _Contract_ sounds very good to me. Put we may found better fields name from weboob data model as you suggest.

**Documentation** : yes, you're right, I will put it there to. I also create a tool to help developers to reuse data : [mesinfosdataplayground](https://github.com/jacquarg/mesinfosdataplayground)

server/konnectors/edf.coffee
jacquarg: Yes it's ugly sometimes. I will have check if I can fix some of them (as the ones you just mention). But some other are so specific, specific to crafts I don't mastery, or related to properties which don't exist out of France, that I fear that the translation may make them meaningless.

server/konnectors/edf.coffee
jacquarg: I think that energy consumption deserve their model. Initiatives like [Green Button](www.greenbuttondata.org/) or Linky in France shows that more and more energy consumption data will come back to the user soon.

server/models/bill.coffee
jacquarg: From the data I receive from EDF, Bill are not directly linked to a contract, but to the client. It may be something typical from the SAP datamodel.
But we could add a contractId field later, if the case occurs.

server/models/bill.coffee
jacquarg: The number is really one of the basic [mandatory fields of a Bill](https://www.service-public.fr/professionnels-entreprises/vosdroits/F31808#fiche-item-1) . 
Number is shorter than vendorBillId ; but if you really think it's better, I can switch to it.

To make it shorter, we could alos use idVendor  (ie bill.idVendor )? Short and explicit ?

server/konnectors/edf.coffee
Phyks: Sure, cozy models can be really specific and contain a lot of data. My concern here is about reusability. It is kind of pointless to have too specific data if it cannot be handled properly by another app. And some data is purely vendor-side data and may not be of any interest for the end user.

Typically, storing the vendor-side identifier of the client inside the Cozy is not useful, as long as we already have a cozy-specific mean of linking items together.

Also, migrating models is a complicated task, requiring to write migrations and handle existing data carefully, so I think introducing new models should be done with extra care :)



Typical example of what I am trying to say above is the new `Contract` model. It might be interesting to have it, and it compares to Weboob `Subscription` (so, it could be reused for any connector relying on Weboob). However, while the Weboob `Subscription` has only 4 specific fields, the `Contract` here has 19 fields. I feel like some of them are vendor-specific (`clientId`, `number`), others are over specific for this particular provider (`numeroDepannage`).

What I would have done is (example for the `Contract` module):
* Define all the models as generic as possible in the cozy models (a `Contract` model as generic as possible, comparable to Weboob `Subscription`).
* Subclass this model, also in the `models` dir, with [use-case](https://github.com/cozy-labs/konnectors/pull/568/files#diff-265233cf1488889e5feae307df30e3b1R16)-specific fields added (typically, this would become an `EnergyProviderContract` or something like this).
* Discard any info too specific to this particular provider, which a general energy provider would not necessarily offer.

This is the same issue with `Bill` vs `InternetBill` vs `PhoneBill` and as far as I know the strategy now is to use `Bill` only.

I am quite new to Cozy and Konnectors, so @m4dz @nicofrand or @ZeHiro will have a better overview of the subject.

server/konnectors/edf.coffee
Phyks: Sure, but energy consumption is nothing more than a number (quantity of energy), a date and a rate (cost per unit amount). I think with these only 3 fields I am able to describe well my energy consumption. But with the same 3 fields, I am also able to describe well any phone call, etc.

server/models/bill.coffee
Phyks: Ok, but I think we should distinguish the data we receive from the service and the data we store. They do not have to match one to one, in my opinion, and we can perform some changes on the data.

In this case, I think the user does not care about which number is associated to him in the SAP. He will most likely not even know what a SAP is. But he may have two contracts (let's say one for his primary house and one for his secondary house) and he wants to know which Bill belongs to which contract. Not sure about the data we get back and if it is doable, but performing changes on the data may be useful.

Same from previous models, if I receive some data from EDF and some data from Enercoop, I do not mind about the provider-specific fields, I just want to compare bills and know how much it costs from each provider.

server/konnectors/edf.coffee
jacquarg: Yes. But, as with Cozy, access control is based on model ; I think it's better to have a different model for phone call log and energy consumption . Because I don't want that apps the app which offers some services around energy have access to my phone call log ; end the phone number of all my correspondents.

server/models/bill.coffee
jacquarg: I have actually not enough information to link Bills and contracts.

jacquarg:
Hi,
I just updated the konnector. It should work now ; only (I hope) the icon is missing. How should I proceed to put the icon (should I edit the sprite ?)
Thanks !

ZeHiro:
Well we (mainteners of konnectors) are still wondering the icons are always generated by the cozy team. @m4dz or @nono can help you on this.

client/app/locales/en.coffee
```diff
@@ -137,6 +137,7 @@ module.exports =
     'konnector description podcast': 'Download your favourite audio podcasts from a RSS feed. This import can take a while.'
     'konnector description materiel_net': 'Import your Materiel.net bills. This konnector requires the Files application to store the bill PDF files.'
     'konnector description meetup': 'Synchronize your Meetup calendar with your Cozy. This konnector requires the Calendar application.'
+    'konnector description edf': "EDF invites you to download your bills, consumption statements, and data about your client's relation,contracts, paiement terms, ... If you use the EDF e.quilibre service, the data will be richer."
```
nicofrand: A comma is missing before contracts, "paiement" should be translated

client/app/locales/fr.coffee
```diff
@@ -138,6 +138,7 @@ module.exports =
     'konnector description podcast': 'Téléchargez vos podcasts audios préférés depuis leurs flux RSS. Les importations peuvent prendre du temps.'
     'konnector description materiel_net': "Télécharge vos factures Materiel.net. Pour pouvoir stocker les factures au format PDF, ce connecteur requiert que l'application Files soit installée sur votre Cozy."
     'konnector description meetup': "Synchronise votre calendrier Meetup sur votre Cozy. Ce connecteur requiert l'application Calendrier."
+    'konnector description edf': "EDF vous propose de télécharger automatiquement vos factures, données de consommation, des données à propos de votre relation client, de vos contrats, modalités de paiement... dans votre Cozy. Pour cela, indiquez l'e-mail et le mot de passe de votre compte client EDF ci-dessous, sélectionnez 'Tous les jours' comme fréquence de mise à jour, puis cliquez sur 'importer et sauvegarder'. (Notez bien que la copie des données que vous aller récupérer est sous votre responsabilitée)"
```
nicofrand: vous aller récupérer est sous votre responsabilitée -> vous allez récupérer est sous votre responsabilité

package.json
```diff
@@ -38,7 +38,9 @@
     "querystring": "0.2.0",
     "request": "2.76.0",
     "request-json": "0.6.1",
-    "ware": "1.3.0"
+    "ware": "1.3.0",
+    "xml2js": "0.4.17"
+
```
nicofrand: Remove useless blank line

package.json
```diff
@@ -154,6 +156,21 @@
     "CozyInstance": {
       "description": "This information is needed to localize the application in your language."
     },
+    "Client": {
+      "description": "Import your Client informations, from the CRM of your vendors."
```
nicofrand: informations -> information

package.json
```diff
@@ -154,6 +156,21 @@
...
+    },
+    "Contract": {
+      "description": "Import informations about your contracts from your vendors."
```
nicofrand: same thing here.

package.json
```diff
@@ -154,6 +156,21 @@
...
+    },
+    "PaymentTerms": {
+      "description": "Import payment modalities from your vendors."
+    },
+    "Home": {
+      "description": "Import some data about your home from EDF."
+    },
+    "ConsumptionStatement": {
+      "description": "Import energy consumption dat ain your Cozy."
```
nicofrand: Please fix the typo

server/lib/update_or_create.js
```diff
@@ -0,0 +1,42 @@
...
+  @param model a cozydb DocType model
+  @param filter a list of field to look at to find similar
+  @param options to be used later.
+*/
+const async = require('async');
+
+module.exports = (log, model, filter, options) =>
+  function(requiredFields, entries, data, next) {
+
+    // function(entries, filter, docType, callback) ->
```
nicofrand: Is this line used ?

server/lib/update_or_create.js
```diff
@@ -0,0 +1,42 @@
...
+      if (err) { return next(err) };
+
+      async.eachSeries(news, function(entry, cb) {
+
+        let toUpdate = docs.find(doc =>
+            filter.reduce((good, k) => good && doc[k] === entry[k], true));
+
+        if (toUpdate) {
+          toUpdate.updateAttributes(entry, cb);
+
```
nicofrand: Useless blank line

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+    return name
+
+edfRequestPost = (path, body, callback) ->
+    K.logger.debug "called edfRequestPost"
+    xmlBody = builder.buildObject body
+    request
+        url: 'https://rce-mobile.edf.com' + path
+        method: 'POST'
+        # TODO : fix SSL3_GET_RECORD:wrong version number
+        #agentOption: securityOptions: 'TLSv1_method'
```
nicofrand: Remove the useless comment please.
And why was SSL v3 chosen here ?

server/lib/base_konnector.js
```diff
@@ -32,6 +34,7 @@ module.exports = {
       slug: slug,
       description: `konnector description ${slug}`,
       logger: logger,
+      docTypeVersion: docTypeVersion,
```
nicofrand: Just a curiosity: why is that needed ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+            'Content-Type': 'application/xml'
+            'Authorization': 'Basic ' +
+                'QUVMTU9CSUxFX0FuZHJvaWRfVjE6QUVMTU9CSUxFX0FuZHJvaWRfVjE='
+            'Accept-Encoding': 'gzip'
+            'Content-Length': xmlBody.length
+        body: xmlBody
+        gzip: true
+    , (err, response, data) ->
+        K.logger.error JSON.stringify(err) if err
+        return callback err if err
```
nicofrand: Please return a localizable key instead of the error message in the callback.

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+        parser.parseString data, (err, result) ->
+            return callback err if err
```
nicofrand: same here

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+                throw null
+
+            if err
+                K.logger.error(
+                    'While fetchEdeliaSimilarHomeYearlyGasComparisions')
+                K.logger.error err
+                throw err
+
+            objs.forEach (obj) ->
+                # if ...
```
nicofrand: Is this if needed ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+        callback null, base64PDF
+
+
+##
+# Edelia
+##
+
+fetchEdeliaToken = (requiredFields, entries, data, callback) ->
+    K.logger.info "fetchEdeliaToken"
+    request.post 'https://api.edelia.fr/' + 'authorization-server/oauth/token',
```
nicofrand: Why separate each string and then concatenate them ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+            paymentTerms.idPayer = getF acoElem, 'ns:numeroPayeur'
+            paymentTerms.payerDivergent = getF acoElem, 'ns:payeurDivergent'
+
+            servicesElem = acoElem['ns:services']
+            services = servicesElem.map (serviceElem) ->
+                service = {}
+                service.name = getF serviceElem, 'ns:nomService'
+                service.status = getF serviceElem, 'ns:etat'
+                service.valeurSouscrite = getF serviceElem, 'ns:valeurSouscrite'
+                service.valeurPossibles = serviceElem['ns:valeursPossibles']
```
nicofrand: Could you translate those properties names ? If not there is a typo I think (a s is missing between valeur and Possibles)

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+            client.homePhone = getF coordonneesElem, 'ns:NumTelFixe'
+            client.email = getF coordonneesElem, 'ns:Email'
+            client.loginEmail = getF coordonneesElem, 'ns:EmailAEL'
+
+            contactElem = getF partenaireElem, 'ns:centreContact'
+            contact = {}
+            contact.title = getF contactElem, 'ns:gsr'
+            contact.phone = getF contactElem, 'ns:telephone'
+
+            adresseElem = getF contactElem, 'ns:adresse'
```
nicofrand: Shouldn't that be "addressElem" ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1368 @@
...
+            'enteteEntree':
+                'ent:jeton': data.edfToken
+
+            'corpsEntree':
+                'numeroBp': entries.client[0].clientId
+
+    edfRequestPost path, body, (err, result) ->
+        return callback err if err
+        try
+            partenaireElem = getF result["ns:msgReponse"], \
```
nicofrand: Should you translate it ?

server/lib/base_konnector.js
jacquarg: It's to help developers who will re-use these data to understand what field they could expect on the data they get from the database of the Cozy. 
The main idea is: the app which create one document write there its name and version, so a future developer want to really know each details about the format, they can go on github and directly check the code which emit it.

server/konnectors/edf.coffee
jacquarg: True, I can remove these line.
Actually, I get sometimes the error `SSL3_GET_RECORD:wrong version number` ; and I didn't manage to really fix it. Actually I am thihnking that's just the EDF endpoint which doesn't respond well sometimes ... : /

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+        try
+            if response.statusCode is 404 or response.statusCode is 500
+                K.logger.warn 'No EdeliaMonthlyElecConsumptions'
+                data.noElec = true
+                throw null
+
+            if err
+                K.logger.error 'Wihle fetchEdeliaMonthlyElecConsumptions'
+                K.logger.error err
+                throw callback err
```
ZeHiro: Why throw here ? Why not just `return callback err`, or just throw err  as the callback is called in the `catch` statement

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+                statement = data.consumptionStatementByYear[obj.year]
+                statement.similarHomes =
+                    site: obj.energies.site
+                    average: obj.energies.similarHomes.SH_AVERAGE_CONSUMING
+                    least: obj.energies.similarHomes.SH_LEAST_CONSUMING
+
+            K.logger.info 'Fetched fetchEdeliaSimilarHomeYearlyElecComparisions'
+        catch e
+            error = e
+        finally
```
ZeHiro: it the finally necessary ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+            callback error
+
+        delete data.consumptionStatementByYear
```
ZeHiro: Unreachable code. Or race condition ad the callback is called inside the finally

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+        error = null
+        try
+            if response.statusCode is 404
+                K.logger.warn 'No EdeliaElecIndexes'
+                throw null
+
+            if err
+                K.logger.error 'Wihle fetchEdeliaElecIndexes'
+                K.logger.error err
+                throw callback err
```
ZeHiro: can you just return ? or throw 'err', as the error is caught in the try catch?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+                    site: obj.energies.site
+                    average: obj.energies.similarHomes.SH_AVERAGE_CONSUMING
+                    least: obj.energies.similarHomes.SH_LEAST_CONSUMING
+
+            K.logger.info 'Fetched fetchEdeliaSimilarHomeYearlyGasComparisions'
+        catch e
+            error = e
+        finally
+            callback error
+        callback()
```
ZeHiro: the callback will be called twice : in the finally statement and here.

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+        error = null
+        try
+            if response.statusCode is 404
+                K.logger.warn 'No EdeliaGasIndexes'
+                throw null
+
+            if err
+                K.logger.error 'Wihle fetchEdeliaGasIndexes'
+                K.logger.error err
+                throw callback err
```
ZeHiro: same as previously

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+
+                statement.statements = statement.statements || []
+                statement.statements.push obj
+
+            K.logger.info 'Fetched fetchEdeliaGasIndexes'
+        catch e
+            error = e
+        finally
+            callback error
+        callback()
```
ZeHiro: double callback call

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+    entries.consumptionstatement = []
+    entries.contract = []
+    entries.bill = []
+    entries.client = []
+    entries.paymentterms = []
+    next()
+
+
+createNewFile = (data, file, callback) ->
+    attachBinary = (newFile) ->
```
ZeHiro: can't you use the File doctType from ../models/file.coffee ?

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+            else
+                callback null, newFile
+
+    # Create file document then attach file stream as binary to that file
+    # document.
+    File.create data, (err, newFile) ->
+        if err
+            callback new Error "Server error while creating file; #{err}"
+        else
+            attachBinary newFile
```
ZeHiro: callback never called.

client/app/locales/fr.coffee
```diff
@@ -138,6 +138,7 @@ module.exports =
     'konnector description podcast': 'Téléchargez vos podcasts audios préférés depuis leurs flux RSS. Les importations peuvent prendre du temps.'
     'konnector description materiel_net': "Télécharge vos factures Materiel.net. Pour pouvoir stocker les factures au format PDF, ce connecteur requiert que l'application Files soit installée sur votre Cozy."
     'konnector description meetup': "Synchronise votre calendrier Meetup sur votre Cozy. Ce connecteur requiert l'application Calendrier."
+    'konnector description edf': "EDF vous propose de télécharger automatiquement vos factures, données de consommation, des données à propos de votre relation client, de vos contrats, modalités de paiement... dans votre Cozy. Pour cela, indiquez l'e-mail et le mot de passe de votre compte client EDF ci-dessous, sélectionnez 'Tous les jours' comme fréquence de mise à jour, puis cliquez sur 'importer et sauvegarder'. (Notez bien que la copie des données que vous allez récupérer est sous votre responsabilité)"
```
ZeHiro: 'Tous les jours' : can't I choose every week/every month?

server/lib/base_konnector.js
```diff
@@ -2,6 +2,7 @@ const _ = require('lodash')
 const printit = require('printit')
 const slugify = require('cozy-slug')
 const fetcher = require('./fetcher')
+const package = require('../../package.json')
```
ZeHiro: Does not compile: 

SyntaxError: server/lib/base_konnector.js: package is a reserved word in strict mode (5:6)
  3 | const slugify = require('cozy-slug')
  4 | const fetcher = require('./fetcher')
> 5 | const package = require('../../package.json')

ZeHiro: Ths path is also wrong. should be ../../../package.json (as the target file will be in build/server/lib...

ZeHiro:
Even after fixing the package.json thing, the init of the application is blocked when initializing the edf konnector

server/konnectors/edf.coffee
```diff
@@ -0,0 +1,1321 @@
...
+    models: [Client, Contract, PaymentTerms, Home, ConsumptionStatement, Bill]
+
+    # Define model requests !
+    init: (callback) ->
+        async.each [Client, Contract, PaymentTerms, Home, ConsumptionStatement]
+        , (docType, cb) ->
+            docType.defineRequest 'all', cozydb.defaultRequests.all, cb
+        , (err) ->
+            if err
+                @logger.error err
```
ZeHiro: missing callback calls: 
```
           if err
               log err
                return callback err
            callback()

ZeHiro:
The import of bills does not seem to work also

server/konnectors/edf.coffee
jacquarg: Yes !! 
Bad refactoring I guess. Thanks !

poupotte:
Tested on https://github.com/jacquarg/konnectors/tree/build and it works.

gregorylegarec:
Thanks @jacquarg !

jacquarg:
Thanks !

server/lib/base_konnector.js
Phyks: Would not it be a good idea to standardize it inside Cozy itself?

jacquarg: That's clearly my goal ; )

PR #577, opened 2016-11-04
Fix test broken on travis

misstick:
It's difficult to maintain konnector environment because of tests that are broken; here is the list of broken test to fix:
 - [x] fix bad tabs,
 - [x] add slug for issue on `captain_train_konnector`,
 - [x] add missing translations,
 - [x] fix typo into tests.

### DIFF
--- a/.transifexrc.tpl
+++ b/.transifexrc.tpl
@@ -1,0 +1,4 @@
+[https://www.transifex.com]
+hostname = https://www.transifex.com
+username = Cozy
+token =
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,33 @@
 language: node_js
 node_js:
-    - 4
+- 4
 services:
-    - couchdb
+- couchdb
 env:
-    global:
-        - NODE_ENV=test
+  global:
+  - NAME=konnectors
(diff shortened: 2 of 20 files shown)


nicofrand:
Hi,

sorry but I don't get it… Why do you disable most of the tests ?

misstick:
Hi @nicofrand; don't worry and take a look at the name of this PR :)
I should add a sentence to explain the checklist into description.

nicofrand:
I did. But I don't understand why you would disable the tests instead of fixing the issues (hence "Don't kill the messenger") ?

ZeHiro:
For konnector migration : https://github.com/cozy-labs/konnectors/blob/master/server/init/konnectors.coffee#L13

misstick:
@ZeHiro : I prefer add a slug property to connector because "guessing" slug by its name is not really serious to my opinion. Anyway, I still use the way to ensure migration and keep consistency.

ZeHiro:
@misstick I think the slug could even be set by konnctors directly from the filename (if possible). As a 2 files with the same filename are not possible in the same directory, I think it will ease the setup.

PR #580, opened 2016-11-08
Add dynamic dialog component

misstick:
- [x] Add basic markup for dialog window,
- [x] Add styles,
- [x] Add close behavior on `closeButton`,
- [x] Add `default` close behavior,
- [x] create a generic component,
- [x] create API (success, error, content etc.),
- <strike>[ ] Add test on this component.</strike>


misstick:
@m4dz : I dont see what should be tested except the markup; should this be done here or later?

m4dz:
Thanks @misstick, I've got some suggestions about how we need to deal whith `dialog` boxes, and your PR point them, nice.

IMO, a dialog can't be handle in a route specificly, as it can't really be predictive (if it's a modal, its a detched head from the rest of the app, and so, it needs to be callable from everywhere). I suggest to rely on `query` params, to define a `d` or `dialog` query parameter that can be called once or multiple times. This way, we can have multiple dialog, in arbitrary URL, when we need.

E.g., we can call it with a `this.$router.push({ query: { d: ['foo', 'bar'] }})` in a component method, so we need here two dialogs : `foo`, and `bar`. Those can e retrieved in the `this.$router.currentRoute.query.d` object (here, it contains an array with `["foo", "bar"]`).

Then, `dialog` boxes should be handled in the `app.vue` component, because they're related to the top level of the app. Using the query params, we can now build a dynamic rendering of the dialog boxes, by creating and invoking a `<k-dialog/>` component: `<k-dialog v-for="dialog in query.d" />` (see [components and v-for](https://vuejs.org/v2/guide/list.html#Components-and-v-for)).

Now we have a custom `<k-dialog/>` component which may be called with some `props`, to determine its type (which kind of content it should load inside), and some low-level API method for closing, validating, etc. It may also trigger some actions that can be catched to react to its internal life-cycle (closing, validating…).

What to you think about this suggestion?

client/app/components/examples/dialog.vue
```diff
@@ -0,0 +1,104 @@
+<template lang="pug">
+    extends ../../templates/layout.pug
+
+    block content
+        include ../../templates/category_list.pug
+
+    block dialog
+        div(v-bind:aria-hidden="hidden", role="dialog")
+            div(role='separator', v-on:click="onClose")
```
m4dz: This one isn't a separator as wanted by the _ARIA_ spec. It's prpobably more a `backdrop` (as in the `<dialog/>` component) but we don't have any `backdrop` role. So I suggest to simply use a `.backdrop` class.

client/app/components/examples/dialog.vue
```diff
@@ -0,0 +1,104 @@
...
+            .wrapper
+                div(role='contentinfo')
+                    header(v-bind:style="headerStyles")
+                        a(v-on:click="onClose", title='close') Close
+                        block dialog_header
```
m4dz: I've got a little doubt about using `block`s, as we probably never need to extend them. I just point it, and let you make a choice.

client/app/components/examples/dialog.vue
```diff
@@ -0,0 +1,104 @@
...
+                },
+                routes: {
+                    success: { name: 'dialogSuccess' },
+                    error: { name: 'dialogError' },
+                    close: { name: 'categoryList' }
+                }
+            }
+        },
+        computed: {
+            headerStyles: {
```
m4dz: 👍

client/app/components/examples/dialog.vue
```diff
@@ -0,0 +1,104 @@
...
+
+<script>
+    export default {
+        data () {
+            return {
+                hidden: false,
+                header: {
+                    'background-image': 'url(test0.png)'
+                },
+                routes: {
```
m4dz: So routes aren't useful anymore according to our discussion.

m4dz:
After a little thinking, we probably just need to handle the closing, by clicking on the `close` button, the backdrop, or by hitting the escape key (see [events key modifiers](https://vuejs.org/v2/guide/events.html#Key-Modifiers)). We probably no need for more API-level actions at this time. 

Whats your thoughs about it?

client/app/app.vue
```diff
@@ -26,7 +26,9 @@
     @import '../node_modules/normalize.css/normalize.css'
     @import './styles/base/_normalize'
     @import './styles/base/_colors'
+
     @import 'cozy-ui'
+    @import './styles/base/_dialog'
```
m4dz: As long as the `_dialog.styl` stylesheet contains rulesets that apply to the dialog component, they should be scoped in the `dialog.vue` file.

client/app/components/category_list.vue
```diff
@@ -1,5 +1,94 @@
 <template lang="pug">
-    main(role='contentinfo')
-        h1 {{ 'my_accounts category title' | t }}
-        article {{ 'my_accounts coming_soon' | t }}
+    extends ../templates/layout.pug
+
+    block content
+        cozy-dialog(v-for="item in dialogs", v-bind:item="item", v-on:success="onSuccess", v-on:error="onError", v-on:submit="onSubmit", v-on:close="onCloseDialog")
```
m4dz: IMHO, we have no need for `onError`, `onSubmit`, `onSuccess` props. Only `onClose` can be useful to hook the closing. For the rest, it's all the concerns of the component that will be embed in the dialog.

Also, I still think the dialog component must be placed in the `app.vue`, cause it's the application responsibility to invoke them from the URL, not the repsonsibility of a sub-component.

client/app/components/category_list.vue
```diff
@@ -1,5 +1,94 @@
...
+            },
+
+
+            onCloseDialog (item) {
+                // Hide <dialog>
+                const index = this.dialogs.indexOf(item)
+                this.dialogs = this.dialogs.splice(index, 0)
+
+                // Update URL query
+                this.$router.push({ query: {}})
```
m4dz: I'm not sure that updating the query should be the responsability of the `onClose` hook. IMO, it's the dialog that updates the query param when invoking its internal `close` method, which occured a rendering of the `app.vue` top component (which listen to query param changes), and then destroy the modal.

client/app/components/dialog.vue
```diff
@@ -0,0 +1,53 @@
+<template lang="pug">
+    div(v-bind:aria-hidden="hidden", role="dialog")
+        div(role='separator', v-on:click="onClose")
+        .wrapper
+            div(role='contentinfo')
```
m4dz: For same reason as invoked in a previous review, this div isn't necessary

client/app/index.js
```diff
@@ -2,6 +2,9 @@ import Vue from 'vue'
 import VueRouter from 'vue-router'
 import VuePolyglot from './plugins/vue-polyglot'
 
+import Dialog from './components/dialog'
+import Notification from './components/notification'
```
m4dz: I understand what you've got in mind with the `Notification` component, but isn't outside of the scope of this PR?

client/app/index.js
```diff
@@ -14,6 +17,10 @@ const router = new VueRouter({ routes })
 
 // Initialize Application
 document.addEventListener('DOMContentLoaded', function initialize () {
+
+  Vue.component('cozy-dialog', Dialog)
```
m4dz: Please, avoid globally declared custom components, it makes the the code complex to understand when trying to know form where comes the components. Prefer the use of the `components` key inside the components that need them (see [local registration](https://vuejs.org/v2/guide/components.html#Local-Registration))

client/app/styles/base/_dialog.styl
```diff
@@ -0,0 +1,16 @@
+[role=dialog]
```
m4dz: All of this ruleset should take place in the `dialog.vue`'s `<style>` section

client/app/templates/category_list.pug
```diff
@@ -0,0 +1,2 @@
+h1 {{ 'my_accounts category title' | t }}
```
m4dz: Why having externalized this as a template?

misstick: it's not usefull anymore but it was before your reviews

client/app/components/dialog.vue
misstick: If I remove this div, it breaks rules defined into `cozy-ui` what should I do? https://github.com/cozy/cozy-ui/blob/master/src/stylus/cozy-ui/components/dialog.styl#L20

client/app/app.vue
```diff
@@ -18,14 +24,100 @@
...
+              const values = query.dialog.split(',')
+              values.forEach((value) => {
+                  this.onOpenDialog(value)
+              })
+          }
+      },
+
+      methods: {
+          onError (err) {
+
```
misstick: @m4dz : Should we use `Vue.config.errorHandler` here or is it a little "too much" ? https://vuejs.org/v2/api/#errorHandler

m4dz: My thoughs about the use of `Vue.config.errorHandler`: it's a smart and clever way to handle and log errors in the stack v2 logging system (poke @nono) when something goes wrong in the app. But to catch some feedback errors (like a misfilled field in a form) I think it's not the best way to do this. Probably having an `$emit` from the component, catched by `app.vue` that pop a notification is a better idea (which can be cumulated with the logging system)

misstick:
We can't add test on components now; `karma` is needed for this and we need to add a `webpack.config` for this.

m4dz:
Congrats 🎉

misstick:
Merry Christmas!

PR #590, opened 2016-11-22
Create missing target folder on import

m4dz:
This PR introduces a way to create missing target folder on import. When an import is triggered, if the target folder doesn't exists (even if it's a nested folder), the full path is created. It's something similar to a `mkdir -p $TARGET_PATH` called in the DS _Folder_ doctype before executing import.


gregorylegarec:
Thanks @m4dz 👍

PR #593, opened 2016-11-23
Mask link to Files

m4dz:
This PR mask the link to the Files folder from the konnector view, when user hadn't completed the onboarding. This way, the link isn't displayed in the embed view in onboarding, it prevents to navigate to an unavailable app from onboarding step.

### DIFF
--- a/client/app/views/konnector.coffee
+++ b/client/app/views/konnector.coffee
@@ -303,7 +303,7 @@ class="folder-link"
 target="_blank">
 #{t "open selected folder"}
 </a>
-"""
+""" if 'ending' in (window.user.onboardedSteps or [])
                 fieldHtml += "</div>"
 
             else
--- a/client/app/views/konnector.coffee
+++ b/client/app/views/konnector.coffee
@@ -303,7 +303,7 @@ class="folder-link"
 target="_blank">
 #{t "open selected folder"}
 </a>
-""" if 'ending' in (window.user.onboardedSteps or [])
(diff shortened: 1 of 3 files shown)


client/app/views/konnector.coffee
```diff
@@ -303,7 +303,7 @@ class="folder-link"
 target="_blank">
 #{t "open selected folder"}
 </a>
-"""
+""" if 'ending' in (window.user.onboardedSteps or [])
```
gregorylegarec: Oops. It seems that I gave you the wrong step name. I am really sorry about that. The good one is `confirmation`. That's the name of the last step. `ending` was the last one in a previous server version.

server/models/user.coffee
```diff
@@ -0,0 +1,14 @@
...
+module.exports = User = americano.getModel 'User',
+    onboardedSteps: Array
+
+
+User.first = (callback) ->
+    User.request 'all', (err, users) ->
+        if err then callback err
+        else if not users or users.length is 0 then callback null, null
+        else
+            user = users[0]
```
gregorylegarec: There is a bug with cozy-db, the `onboardedSteps` property returned by cozy-db is an array of array instead of an array.

We handled it in cozy-proxy here https://github.com/cozy/cozy-proxy/blob/development/server/models/user.coffee#L83 and here https://github.com/cozy/cozy-proxy/blob/development/server/models/user.coffee#L28

gregorylegarec:
Thanks @m4dz ! 👍

PR #595, opened 2016-11-24
Official Maif Konnector (MesInfos)

jacquarg:
This is the konnector for Maif data. It has been realised by developers from Maif (see https://github.com/cozy-MAIF/konnectors). But we haven't been explicit enough about the procedure of integration of the konnectors and so, to save some time, I am doing this integration for them.

I know there is still work to do on it:

- [x] fix lint
- [x] icon (already pinged @kelukelu ) #597 

There is also an issue about data model, code could be simplified (request-json ...), or write in a more konnector way. But there is already one app using this data as is (https://github.com/cozy-MAIF/mes-infos-maif), and we are very close a big milestone ; so I propose to work on that later, with Maif and you, simultaneously on their app and the konnector.

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -137,6 +137,7 @@ module.exports =
     'konnector description podcast': 'Download your favourite audio podcasts from a RSS feed. This import can take a while.'
     'konnector description materiel_net': 'Import your Materiel.net bills. This konnector requires the Files application to store the bill PDF files.'
     'konnector description meetup': 'Synchronize your Meetup calendar with your Cozy. This konnector requires the Calendar application.'
+    'konnector description maif': "Maif invites you to download some data in your Cozy (societary, home, claims, ...). To do that, (1) use the 'connect' button to connect on your Maif account, (2) select 'every day' as update schedule, then (3) click on 'Import and Save'."
 
     # Konnectors' notifications
     'notification import error': 'an error occurred during import of data'
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -139,7 +139,8 @@ module.exports =
     'konnector description podcast': 'Download your favourite audio podcasts from a RSS feed. This import can take a while.'
     'konnector description materiel_net': 'Import your Materiel.net bills. This konnector requires the Files application to store the bill PDF files.'
     'konnector description meetup': 'Synchronize your Meetup calendar with your Cozy. This konnector requires the Calendar application.'
-    'konnector description maif': "Maif invites you to download some data in your Cozy (societary, home, claims, ...). To do that, (1) use the 'connect' button to connect on your Maif account, (2) select 'every day' as update schedule, then (3) click on 'Import and Save'."
+    'konnector description maif': "Maif invites you to download some data in your Cozy (societary, home, claims, ...)."
+    'konnector customview maif': "To do that<ol><li>use the 'connect' button to connect on your Maif account,</li><li>select 'every day' as update schedule,</li></li>click on 'Import and Save'."
 
     # Konnectors' notifications
     'notification import error': 'an error occurred during import of data'
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -44,6 +44,8 @@ module.exports =
     'loginUrl': "Login URL"
     'token': "Token"
     'refreshToken': "Refresh Token"
+    'code': "Authorization code"
+    'redirectPath': "Path of the redirect URL"
 
     'home headline': """
     With Konnectors you can retrieve many data and save them into your Cozy.
--- a/client/app/locales/fr.coffee
+++ b/client/app/locales/fr.coffee
@@ -28,7 +28,7 @@ module.exports =
     'apikey': 'Api key'
     'phoneNumber': 'Numéro de téléphone'
     'folderPath': 'Chemin du dossier'
-    'none' : 'Aucun'
+    'none' : 'Pas de mise à jour auto'
     'every hour': 'Toutes les heures'
     'every day': 'Tous les jours'
     'every week': 'Toutes les semaines'
@@ -138,7 +138,7 @@ module.exports =
     'konnector description podcast': 'Téléchargez vos podcasts audios préférés depuis leurs flux RSS. Les importations peuvent prendre du temps.'
     'konnector description materiel_net': "Télécharge vos factures Materiel.net. Pour pouvoir stocker les factures au format PDF, ce connecteur requiert que l'application Files soit installée sur votre Cozy."
     'konnector description meetup': "Synchronise votre calendrier Meetup sur votre Cozy. Ce connecteur requiert l'application Calendrier."
-
+    'konnector description maif': "La Maif vous propose de télécharger automatiquement des données vous concernant (sociétaire, foyer, véhicule, sinistre, ...) dans votre Cozy. Pour cela (1) connectez-vous à votre compte en ligne Maif via le bouton 'connect', (2) sélectionnez 'Tous les jours' comme fréquence de mise à jour, puis (3) cliquez sur 'importer et sauvegarder'."
     # Konnectors' notifications
     'notification import error': "une erreur est survenue pendant l'importation des données"
     'notification prefix': "Konnector %{name} :"
--- a/client/app/locales/fr.coffee
+++ b/client/app/locales/fr.coffee
@@ -44,6 +44,9 @@ module.exports =
     'loginUrl': "URL d'authentification"
     'token': "Token"
     'refreshToken': "Mettre à jour le token"
+    'code': "Code d'autorisation"
+    'redirectPath': "Chemin de l'URL de redirection"
+
 
     'home headline': """
     Konnectors vous permet de récupérer de nombreuses données et de les intégrer à votre Cozy.
--- a/client/app/locales/fr.coffee
+++ b/client/app/locales/fr.coffee
@@ -141,7 +141,8 @@ module.exports =
     'konnector description podcast': 'Téléchargez vos podcasts audios préférés depuis leurs flux RSS. Les importations peuvent prendre du temps.'
     'konnector description materiel_net': "Télécharge vos factures Materiel.net. Pour pouvoir stocker les factures au format PDF, ce connecteur requiert que l'application Files soit installée sur votre Cozy."
     'konnector description meetup': "Synchronise votre calendrier Meetup sur votre Cozy. Ce connecteur requiert l'application Calendrier."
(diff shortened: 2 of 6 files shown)


qqOBq
```diff
@@ -0,0 +1,84 @@
...
+[1m--- a/client/app/locales/en.coffee[m
+[1m+++ b/client/app/locales/en.coffee[m
+[36m@@ -137,6 +137,7 @@[m [mmodule.exports =[m
+     'konnector description podcast': 'Download your favourite audio podcasts from a RSS feed. This import can take a while.'[m
+     'konnector description materiel_net': 'Import your Materiel.net bills. This konnector requires the Files application to store the bill PDF files.'[m
+     'konnector description meetup': 'Synchronize your Meetup calendar with your Cozy. This konnector requires the Calendar application.'[m
+[32m+[m[32m    'konnector description maif': "Maif invites you to download some data in your Cozy (societary, home, claims, ...). To do that, (1) use the 'connect' button to connect on your Maif account, (2) select 'every day' as update schedule, then (3) click on 'Import and Save'."[m
+ [m
+     # Konnectors' notifications[m
+     'notification import error': 'an error occurred during import of data'[m
```
ZeHiro: What is this file?

jacquarg: Very good question ! 
I m gona clean this right now. Thanks !

server/konnectors/maif.js
```diff
@@ -0,0 +1,313 @@
...
+      app: appToOpen,
+    }
+  });
+}
+
+/**
+* generate UUID for nonce and state parameters
+*/
+function generateUUID() {
+  let d = new Date().getTime();
```
nicofrand: Please use Date.now()

server/konnectors/maif.js
```diff
@@ -0,0 +1,313 @@
...
+  const uuid = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
+    const r = (d + Math.random() * 16) % 16 | 0;
```
nicofrand: What are you trying to achieve here ?
If you just want to cast the float into a number, can you use Math.floor instead ?

server/konnectors/maif.js
```diff
@@ -0,0 +1,313 @@
...
+    d = Math.floor(d / 16);
+    return (c === 'x' ? r : (r & 0x3 | 0x8)).toString(16);
```
nicofrand: Looks like what you are using is http://stackoverflow.com/questions/24203203/understanding-javascript-function-in-base-16.

Could you use http://stackoverflow.com/a/24203297 instead ? (I did not test)

jacquarg: Thank's, I didn't notice it was a s-o contribution ^^.
It may mean that we don't need a uuid construct in this way, but just some uuid. So I will make some test with some cleaner and more integrated solutions.
I make a proposition soun.

ZeHiro:
@jacquard i'll try the konnector as soon as my computer iss repaired. I have an account at maif.

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+    }
+  });
+}
+
+/**
+* Display a notification in Cozy
+* code : code of the message to send
+* appToOpen : Link to the app that will be opened on notification's click
+*/
+function sendNotification(code, appToOpen) {
```
ZeHiro: When I click on "Connect" on the maif page I read : 'En validant votre identification, Cozy Cloud, recevra les informations suivants :' I would expect : 'En validant votre identification, les informations suiantes seront importées dans votre Cozy Cloud' :

ZeHiro: I would expect that clicking on 'Connect' opens a new tab. can you add `target="_blank"` here : https://github.com/cozy-labs/konnectors/blob/master/client/app/views/konnector.coffee#L330

client/app/locales/en.coffee
```diff
@@ -137,6 +139,7 @@ module.exports =
     'konnector description podcast': 'Download your favourite audio podcasts from a RSS feed. This import can take a while.'
     'konnector description materiel_net': 'Import your Materiel.net bills. This konnector requires the Files application to store the bill PDF files.'
     'konnector description meetup': 'Synchronize your Meetup calendar with your Cozy. This konnector requires the Calendar application.'
+    'konnector description maif': "Maif invites you to download some data in your Cozy (societary, home, claims, ...). To do that, (1) use the 'connect' button to connect on your Maif account, (2) select 'every day' as update schedule, then (3) click on 'Import and Save'."
```
ZeHiro: To have a nicer display, you could use the customView attribute of the maif konnector, and use a `<ol>`. Why is every day required? I would like to do it weekly, is it possible? If it is please update the label with `choose the update freqncy` or so.

client/app/locales/fr.coffee
```diff
@@ -138,7 +141,7 @@ module.exports =
     'konnector description podcast': 'Téléchargez vos podcasts audios préférés depuis leurs flux RSS. Les importations peuvent prendre du temps.'
     'konnector description materiel_net': "Télécharge vos factures Materiel.net. Pour pouvoir stocker les factures au format PDF, ce connecteur requiert que l'application Files soit installée sur votre Cozy."
     'konnector description meetup': "Synchronise votre calendrier Meetup sur votre Cozy. Ce connecteur requiert l'application Calendrier."
-
+    'konnector description maif': "La Maif vous propose de télécharger automatiquement des données vous concernant (sociétaire, foyer, véhicule, sinistre, ...) dans votre Cozy. Pour cela (1) connectez-vous à votre compte en ligne Maif via le bouton 'connect', (2) sélectionnez 'Tous les jours' comme fréquence de mise à jour, puis (3) cliquez sur 'importer et sauvegarder'."
```
ZeHiro: same as for en.coffee

server/controllers/konnectors.coffee
```diff
@@ -106,6 +106,10 @@ module.exports =
             for k, v of req.query
                 account[k] = v
 
+            # Add redirection path, used by some konnector to build back the
+            # redirectUri
+            account.redirectPath = req.originalUrl
```
ZeHiro: can you add a check that req.originalIrl does exist ?

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+const env = 'prod'; // dev / pprod / prod
+
+let connectUrl;
+let apikey;
+let infoUrl;
+let clientId;
+let secret;
+
+switch (env) {
+  case 'dev':
```
ZeHiro: can you leave only the prod seting ?

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+  state = uuid();
+}
+
+if (nonce === '') {
+  nonce = uuid();
+}
+
+const MaifUser = cozydb.getModel('MaifUser', {
+  password: String,
+  profile: Object,
```
ZeHiro: Can you provide in comment what object this is supposed to be?

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+  date: String,
+});
+
+const connecteur = module.exports = factory.createNew({
+  name: 'MAIF',
+  slug: 'maif',
```
ZeHiro: unnecessary as createNew will force the slug

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+      let jsonToken = null;
+      try {
+        jsonToken = JSON.parse(body);
+      } catch (e) {
+        err = 'parsing error';
+      }
+
+      if (err != null) {
+        connecteur.logger.error(err);
+        callback('Erreur lors de la récupération des données.');
```
ZeHiro: can you use `callback(err)` instead ?

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+* refreshToken function
+* called at each scheduled import (every hour/day/week/month)
+* get new token and refresh token
+* call getToken with token and refresh token
+*/
+function refreshToken(requiredFields, entries, data, next) {
+  MaifUser.first((err, maifUser) => {
+    let tokenValid = true;
```
ZeHiro: can you add a check over err ? 
`if err then return next(err)`

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+          headers: {
+            Authorization: `Basic ${b64Client}`,
+          },
+          form: {
+            grant_type: 'refresh_token',
+            refresh_token: token,
+          }
+        };
+        request(options, (err, response, body) => {
+          try {
```
ZeHiro: can you add a check over the err ? + return callback('request error')

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+      url: infoUrl,
+      jar: true,
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${token}`,
+      }
+    };
+
+    request(options, (err, response, body) => {
+      try {
```
ZeHiro: can you add a check over the err ? + return callback('request error')

server/konnectors/maif.js
```diff
@@ -0,0 +1,302 @@
...
+        grant_type: 'authorization_code',
+        code: requiredFields.code,
+        state,
+        nonce,
+        redirect_uri: urlRedirect
+      }
+    };
+    connecteur.logger.info(options);
+    request(options, (err, response, body) => {
+      let jsonToken = null;
```
ZeHiro: can you add a check over the err ? + return callback('request error')

server/konnectors/maif.js
jacquarg: I have no control about the labels on the maif page. But I totally agree with your proposition, and will send it to Maif.

About the new tabs. It's a bit more complicated: at the end of the Maif connection, there is a redirection back to the konnector ; and if it's not back to the original tab, it's very confusing.
We should think about it, maybe use an other iframe, maybe a pop-up, for the newt version of my account ?

client/app/locales/en.coffee
jacquarg: Ok for the ol, it will be cleaner.
We advice daily update because update on the data could happen, and we which that the user could look into it's Cozy to view them (not ve forced to go back to some other Maif interface to view the uptodate informations). But, as the user get more experienced, he will understand that he have the control on this setting.

server/konnectors/maif.js
```diff
@@ -206,14 +208,10 @@ function getData(token, callback) {
 */
 function refreshToken(requiredFields, entries, data, next) {
   MaifUser.first((err, maifUser) => {
-    if (err) {
```
ZeHiro: why do you remove the callback err here ?

poupotte:
It's ok for me !

server/konnectors/maif.js
jacquarg: I removed it because here, we use the err as a signal that the konnector is not configured yet ; and that we need to go fetch the token. 
This is badly designed, but didn't want to rewrite all the code from maif in a hurry, and without them.

ZeHiro:
@poupotte I merge then? (looks like you don't have the right for it)

poupotte:
Yes, I can't merge it but I tested it and it works :)

poupotte:
Thanks @ZeHiro !

jacquarg:
Thank's  !

PR #601, opened 2016-11-28
Add konnectors list on categories view (without sorting) using a generic list item component

CPatchane:
Thanks @GoOz to have a review on this pull request.

This PR:
* [x] Change the config webpack to load also all assets in `assets/icons/`
* [x] Add new connectors icons
* [x] Add a generic list item component to be used for konnectors and use cases lists (/poke @m4dz)
* [x] Add documentation related to this component
* [x] Add the non-connected konnectors list on the categories view (without categories sorting)
* [x] Add the item component tests
* [x] Add new locales


client/app/components/list_item.vue
```diff
@@ -0,0 +1,106 @@
...
+            transition          .2s ease all
+
+        header
+            border-radius       4px 4px 0px 0px
+            height              9em
+            position            relative
+
+            .item-icon
+                border-radius   4px 4px 0px 0px
+                width           14em
```
m4dz: 👍

client/test/components/list_item.spec.js
```diff
@@ -0,0 +1,127 @@
...
+  listItem.props.title = 'A component title'
+  listItem.props.link = '#'
+
+
+  afterEach(function() {
+    if (vm && vm._isMounted) vm.$destroy()
+  })
+
+
+  describe('data property', () => {
```
m4dz: Are testing datas really useful?

CPatchane: Here we have data that should work both if we provide a backgroundCSS value and if not, so I thought useful to test it. But maybe it's discussable. What do you think?

client/app/components/category_list.vue
```diff
@@ -3,19 +3,42 @@
...
+
     export default {
-      methods: {
-          openDialog () {
-              this.$emit('open-dialog', 'dialog-1')
-          }
-      }
+        data: function () {
+            return {
+                konnectorsList: window.initKonnectors
```
CPatchane: We shouldn't use the list from the `window` directly inside the component and externalise it or we shouldn't use it at all? For the first case I can put it in an helper for better reusability.

GoOz:
Looks good to me, as far as i can tell. Semantic and style seem appropriate in this case.

client/app/components/category_list.vue
m4dz: my main concern is really about testing: we shouldn't rely on external resources, so having them wrappeed in a helper that can be easily mocked is probably a better solution. But this code will be converted, so don't worry, it was just a remark, no a real need for this PR :)

PR #602, opened 2016-11-28
Fix magic folder autoselect

m4dz:
This PR introduces some changes in the magic folder for a konnector:
- adds the magic folder at end and select it if the konector isn't configured and the folder doesn't exists
- select the magic folder into the list if it already exists and the konnector isn't configured
- select the right folder (the selected one or the magic folder when created) when the konnector is configured

It should fix the select value issue.


gregorylegarec:
Thanks @m4dz.

PR #605, opened 2016-11-30
[tech] Migration to preact

goldoraf:
This PR should give us a preact codebase roughly equivalent to the existing vue.js one. It's still lacking tests, I'm currently working on integrating enzyme with preact so that we could properly unit test our components.


client/package.json
```diff
@@ -2,14 +2,15 @@
   "dependencies": {
     "node-polyglot": "2.0.0",
     "normalize.css": "5.0.0",
-    "vue": "^2.1.0",
-    "vue-resource": "1.0.3",
-    "vue-router": "2.0.1"
+    "preact": "^6.4.0",
+    "preact-compat": "^3.9.2",
```
CPatchane: For my culture, as we want to use directly `preact` here, we still need the `preact-compat` dependance?

client/app/app.js
```diff
@@ -0,0 +1,51 @@
...
+                                <ConnectedList path="/connected"/>
+                            </Router>
+                        </div>
+                    </main>
+                    {selectedItem && <ItemDialog item={selectedItem} onClose={this.onItemUnselect} />}
+                </div>
+            </I18n>
+        )
+    }
+}
```
CPatchane: I noticed that all your new files don't have the last trailing empty line, I think we should keep them as this is also expected from the cozy-guidlines. What do you think about?

client/app/app.js
```diff
@@ -0,0 +1,51 @@
...
+import ItemDialog from './components/item_dialog'
+
+export default class App extends Component {
+    constructor(props) {
+        super(props)
+        this.state = {
+            konnectors: props.konnectors,
+            selectedItem: null
+        }
+        this.onItemSelect = this.onItemSelect.bind(this)
```
CPatchane: We still use this classic binding way or we plan to use a decorator for this? This way is good for me, it's just to keep coherent. Or maybe we can use the decorator later if necessary.

client/app/styles/index.styl
```diff
@@ -0,0 +1,111 @@
...
+        text-decoration: none
+        color: $red
+        flex: 1
+        padding: 1em 1.5em
+        margin: 0.25em 0
+
+        &:hover:not(.router-link-active)
+            background-color: $grey-01-alpha
+
+    .router-link-active
```
CPatchane: I tried your branch and I figured out that this style is not applied anymore, we don't have the red highlight in the menu anymore. That part was specific to the previous [vue-router](https://router.vuejs.org/en/api/router-link.html). However, the `preact-router` seems to not work like the react-router with a `activeClassName` attribute.

client/app/app.js
```diff
@@ -0,0 +1,51 @@
+import { h, Component } from 'preact'
```
CPatchane: Sorry for this double review but I didn't noticed it before. I think we should use `.jsx` extension for files handling JSX code, it's better for linting and highlighting when we develop using this language.

client/app/components/item_dialog.js
```diff
@@ -0,0 +1,25 @@
...
+                    <main>
+                        <p>Foo</p>
+                    </main>
+                    <footer>
+
+                    </footer>
+                </header>
+            </div>
+        </div>
+        <h1>{t('my_accounts connected title')}</h1>
```
CPatchane: Oops, that seems be here by mistake from a copy/paste action..

client/package.json
goldoraf: If we want to use React libs like react-router and/or stay compatible with other JSX-compatible lib like inferno, it'd be better to use preact-compat and rewrite our imports for react.

client/app/app.js
goldoraf: I thought we could always use the decorator later, or that we could use a better router to handle that use case.

client/app/styles/index.styl
goldoraf: Fixed using react-router.

client/app/app.js
goldoraf: I handle all JS files as JSX files in Sublime. I'd prefer to stay with the standard .js, but I'm not closed to the other option. Sure of that ?

CPatchane: It's a personal preference to really separate the JSX part from the pure JS. And we can have `a .jsx file = a (p)react component`. But maybe we can have others opinion. @gregorylegarec ? @m4dz ? or @GoOz ?

gregorylegarec: Personally, I prefer to have a file extension that describes what is in the file. If I got a .js, file I am not expecting to find some JSX or other stuff inside.
So my preference goes to separated .js and .jsx file.
However I may be missing a point.

client/app/app.js
gregorylegarec: Sorry I did not see this comment yesterday. I think that we have to make a choice and keep to it. If it's in the Cozy guidelines and it's already the case in other project, let's keep it.

(To be honest, I use to add a empty line at the end of all my files but I don't really know what role has this empty line.)

CPatchane: @GoOz Same thing for Atom as well

kosssi: it's better with, it's a standard directly from C:
- https://robots.thoughtbot.com/no-newline-at-end-of-file

But it's for github no warning too ;)

@goldoraf you can automate it on your IDE

CPatchane:
Thanks and great job @goldoraf !

PR #607, opened 2016-12-01
Improve EDF konnector quality

jacquarg:
Fix an possible future bug on cozydb request missing, by moving models to server/models.
Improve fetch success, adding a retry.


ZeHiro:
@jacquarg can you also use a string here, instead of an instance of error : 
https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/edf.coffee#L147

and here : 
https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/edf.coffee#L1175

The error is not caught here : 

https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/edf.coffee#L1195

poupotte:
It works for me :)

PR #610, opened 2016-12-01
Use cases and other improvements

CPatchane:
This PR:

* [x] Fix the default icon behaviour in the account_item component (to not display default icon when asked, for use cases for example)
* [x] Rename index.js to index.jsx
* [x] Add filters for categories (only no connected accounts) and connected (only connected accounts) views
* [x] Add use cases list in the discovery view
* [x] Add a redirection for '/' to redirect to '/discovery'
* [x] Add modal behaviour for use cases items and connected accounts item via the Router
* [x] Fix some styles


CPatchane:
@goldoraf For the naming convention, I prefer using camelCase in général in JavaScript. Better again if it's the convention for the react ecosystem. I also agree with you to not mix the two ways here.

goldoraf:
Thanks a lot!

PR #616, opened 2016-12-06
Fix use cases image and title translation

CPatchane:
This PR:

* [x] Fix a scroll weird behaviour for the aside bar
* [x] Fix image paths for use cases item
* [x] Add translation for use cases item title


m4dz:
Nice shoot :)

PR #618, opened 2016-12-06
[debug] Fix SVG color in navigation's item when selected on Chrome

GoOz:
Fill attributes on path in SVGs prevent CSS from changing the color, unless you specify the target to the path of the SVG (instead of the SVG itself) but Chrome cannot change path's color.

So i removed the fill atrributes, which prevented the use of those SVGs with any other color anywhere else on the app anyway, and specified the colors through CSS.


m4dz:
Thanks 🎉

PR #619, opened 2016-12-07
Add use case dialogs content + some generic dialog component styles

CPatchane:
This PR:

* [x] Remove the reference to "konnector" in the use case index.json (with changes in doc) to use "account"
* [x] Add the content of the use case dialog with the related accounts list
* [x] Add sub-dialog for accounts in the use case dialog
* [x] Add styles for the use case dialog
* [x] Add generic close button (X)
* [x] Add some generic styles (also for the account dialog)

### DIFF
--- a/client/app/components/AccountDialog.jsx
+++ b/client/app/components/AccountDialog.jsx
@@ -25,7 +25,10 @@ const AccountDialog = ({ t, item, iconName, enableDefaultIcon }) => (
   <div role='dialog' class='account-dialog'>
     <div class='wrapper'>
       <div role='contentinfo'>
-        <header style={{background: item.color.css || 'white'}}>
+        <header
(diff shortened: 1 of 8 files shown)


client/app/components/AccountDialog.jsx
```diff
@@ -4,21 +4,42 @@ import { translate } from '../plugins/preact-polyglot'
 import { withRouter } from 'react-router'
 
 const CloseButton = withRouter(({ router }) => (
-  <button role='close' onClick={router.goBack}>Close</button>
+  <div class='close-button' role='close' onClick={router.goBack} />
```
goldoraf: Using a button is more semantic in my eyes

goldoraf:
Very good job!

PR #621, opened 2016-12-08
Update all accounts fields property

CPatchane:
Update all accounts fields property to use object format instead of simple strings


goldoraf:
Looking great, thanks a lot!

PR #624, opened 2016-12-12
Account initial configuration

goldoraf:
There is still some work to do: connecting the new store to the views that depend on it, adding some tests, ... But as it's quite a large PR, the sooner the review starts, the better :)


client/app/lib/statefulForm.js
```diff
@@ -0,0 +1,127 @@
+/** @jsx h */
+import { h, Component } from 'preact'
```
CPatchane: What do you think to use the `.jsx` extension here? Even if it's more an helper than a component.

client/app/styles/index.styl
```diff
@@ -179,3 +293,56 @@ aside
                 height      6em
                 width       100%
                 margin      2em 0em
+
+// ONBOARDING NOTIFICATION
```
CPatchane: This comment should be updated for this app, there is no relation to the cozy-proxy onboarding here.

client/app/components/AccountConfigForm.jsx
```diff
@@ -0,0 +1,37 @@
...
+    fields,
+    values
+  }
+}
+
+const AccountConfigForm = ({ t, fields, dirty, submit, submitting }) => (
+  <div class='account-form'>
+    {mapObject(fields, (field, name) => <Field label={t(name)} {...field} />)}
+    <div class='account-form-controls'>
+      <button aria-busy={submitting ? 'true' : 'false'} onClick={submit}>{t('my_accounts account config button')}</button>
```
CPatchane: We could indent the button content here.

client/app/components/Field.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Field = (props) => {
+  let inputs
+  if (props.children.length !== 0) {
+    inputs = props.children.map(
+      child => cloneElement(child, Object.assign(props, {selected: props.value, className: 'account-field-input'}))
+    )
+  } else {
+    const { type, placeholder, value, onChange, onBlur } = props
+    inputs = (<input type={type} placeholder={placeholder} className='account-field-input' value={value} onChange={onChange} onBlur={onBlur} />)
```
CPatchane: This line seems to be too long.

client/app/components/Field.jsx
```diff
@@ -0,0 +1,37 @@
+/** @jsx h */
+import { h, cloneElement } from 'preact'
+import classNames from 'classnames'
+
+const Field = (props) => {
+  let inputs
+  if (props.children.length !== 0) {
+    inputs = props.children.map(
+      child => cloneElement(child, Object.assign(props, {selected: props.value, className: 'account-field-input'}))
```
CPatchane: Splittable.

client/app/components/Field.jsx
```diff
@@ -0,0 +1,37 @@
...
+  var classes = classNames('account-field', {
+    'account-field--required': required === true,
+    'account-field--error': errors.length !== 0,
+    'account-field--dirty': dirty === true || touched === true
+  })
+  return (
+    <div className={classes}>
+      <label>{label}</label>
+      {children}
+      {errors.length !== 0 && errors.map((err, i) => <small key={i} className='account-field-error'>{err}</small>)}
```
CPatchane: Splittable.

client/app/components/Notifier.jsx
```diff
@@ -0,0 +1,106 @@
...
+        ...this.state.notifications.slice(idx + 1)
+      ]
+    })
+  }
+
+  render () {
+    return (
+      <div className='coz-notifier'>
+        {this.state.notifications.map(notif => (
+          <Notification type={notif.type} message={notif.msg} onClose={this.handleClose.bind(this, notif.id)} />
```
CPatchane: Splittable.

client/app/lib/accountStore.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+export const connectToStore = (mapStateToProps, mapStoreToProps) => {
+  return (WrappedComponent) => {
+    class Connected extends Component {
+      constructor (props, context) {
+        super(props, context)
+        this.store = context.store
+        this.state = Object.assign({}, mapStateToProps(this.store.getState()), mapStoreToProps(this.store, props))
```
CPatchane: Splittable.

client/app/styles/index.styl
```diff
@@ -179,3 +293,56 @@ aside
...
+
+    box-sizing border-box
+    padding 1em
+
+    color white
+
+    // When only a title is displayed
+    min-height: (96/base-font-size)em
+    padding-left (96/base-font-size)em
+    background-image url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='48' height='48' viewBox='0 0 48 48'><path fill='#FFFFFF' fill-rule='evenodd' d='M24,48 C37.254834,48 48,37.254834 48,24 C48,10.745166 37.254834,0 24,0 C10.745166,0 0,10.745166 0,24 C0,37.254834 10.745166,48 24,48 Z M13.4142136,24.5857864 C12.633165,23.8047379 11.366835,23.8047379 10.5857864,24.5857864 C9.80473785,25.366835 9.80473785,26.633165 10.5857864,27.4142136 L17.5857864,34.4142136 C18.366835,35.1952621 19.633165,35.1952621 20.4142136,34.4142136 L35.4142136,19.4142136 C36.1952621,18.633165 36.1952621,17.366835 35.4142136,16.5857864 C34.633165,15.8047379 33.366835,15.8047379 32.5857864,16.5857864 L19,30.1715729 L13.4142136,24.5857864 Z'/></svg>")
```
CPatchane: Why not use external `.svg`? I think it's better to keep separate the styles and the assets. We already have discussed about this with @gregorylegarec, so I just give my opinion about that here. What do you think? Any opinion @GoOz or @m4dz ?

GoOz: Well, not that svg in base64 is a bad idea itself, but it'll be better if we all stick to the same way we treat SVGs in CSS backgrounds. 
The way i see it, external SVGs are better for maintenance but base64 SVGs can be better for performance in certain cases and not so good in some other cases :/ So from my point of view, external SVGs are the best for the job as it sucks in less situations than base64 SVGs do.
If external assets it is then yeah, it should be changed.
My 2 cents.

client/app/components/AccountConfigForm.jsx
```diff
@@ -0,0 +1,37 @@
...
+import Field from './Field'
+
+const mapObject = (obj, cb) => Object.keys(obj).map(k => cb(obj[k], k))
+
+const formConfig = ({ t, fields, slug }) => {
+  let values = {}
+  Object.keys(fields).forEach(name => {
+    if (fields[name].advanced) {
+      fields[name].type = 'hidden'
+      if (fields[name].default) {
```
goldoraf: Yes of course :)

client/app/components/Notifier.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+  notify (notification) {
+    this.setState({
+      notifications: [...this.state.notifications, notification]
+    })
+  }
+
+  handleClose (id) {
+    let idx
+    this.state.notifications.forEach((n, i) => { if (n.id === id) idx = i }) // le find du pauvre...
```
goldoraf: yeah, why not? :)

client/app/lib/statefulForm.js
goldoraf: Yes, it's more of a helper, but still it contains JSX... Not sure what to do here...

CPatchane: I just think if there is some JSX, we can use `.jsx`. And we could have helper/library in `.jsx`?

client/app/components/Notifier.jsx
CPatchane: I think it's better to use polyfills if/when needed. We could use this function anywhere else in the app. After discussions with @gregorylegarec, polyfills can be handled by a simple webpack plugin, so that could be our simple compatibilities solution here.

client/app/styles/index.styl
goldoraf: I just used the styles @gregorylegarec wrote for cozy-proxy, using base64 SVGs is his idea :)

m4dz: I definitely enforce the use of SVG embed in stylesheet, especially for icons (and I'll be a nazi on this, you're now aware :)).

Let's say your on a mobile connectivity, you CSS is loaded but the ruleset containing the external SVG isn't applied at this time because it concerns a notification that'll pop later. Then suddenly (like our [Francis](http://www.francisblaireau.com/) friend) you connection drop, and your notification pop. You now face an element without its icon because this one can't be retrieved as there's not connectivity at all. It sucks. A lot. Really.

With embeded SVG icons, when your CSS is applied it means the stylesheet is fully loaded. So icons are available and even if your connection drop (Hop!), you'll never loose the Ui.

But I agree embedding SVG in stylesheet is a bit messy, this is why stylus provides us the magnificent [`embedurl`](http://stylus-lang.com/docs/functions.url.html#utf8-encoding-for-svgs) directive, so please use it to embed external source SVG files in the final CSS. I'll bring you kisses.

Thanks :)

Your devoted Ui Nazi

m4dz:
@CPatchane about line length, I'm for concise code. JSX brings XML syntax in our files, so forcing us to limit length will be counter-productive IMHO. I suggest we write JX properly and readably in the more concise way as we can, and count on common-sense for the rest of our code ;)

client/app/styles/index.styl
GoOz: @m4dz True. A tool to generate embed SVG would fix the maintenance issue but i must ask, if you have a lot of very complex SVGs with multiple complex paths (which is not the case here but still), doesn't the size of your CSS critically increase? Or is it not so relevant in terms of performance?

CPatchane:
Very nice job @goldoraf !

client/app/styles/index.styl
m4dz: As discussing out of this thread, complex SVG are often for illustration rather than icons. So we can stuck on external files in this case.

Another case is to pass SVG through an optimizer like SVGO, which can be handled by webpack. I add it to our `cozy-template` skeleton :)

PR #626, opened 2016-12-13
Better default values

CPatchane:
* [x] Use <account> in default values fallbacks
* [x] Use the account name instead of the slug for <account> variable replacement


goldoraf:
:+1:

PR #628, opened 2016-12-15
Account configuration fixes from early feedback

goldoraf:


client/app/components/AccountConnectionForm.jsx
```diff
@@ -3,14 +3,14 @@ import { h } from 'preact'
...
+import Field, { PasswordField } from './Field'
 
-const formConfig = ({ t, fields, accountName }) => {
+const formConfig = ({ t, fields, slug }) => {
   let values = {}
   Object.keys(fields).forEach(name => {
     if (fields[name].default) {
       values[name] = fields[name].default.replace(/<my_accounts>/gi, t('my_accounts title'))
-                                         .replace(/<account>/gi, accountName)
+                                         .replace(/<account>/gi, slug)
```
CPatchane: The \<account\> is supposed to be the account __name__, __not the slug__. I replaced that in my last PR. Is there an issue somewhere else which expect the slug instead of the name? Why is it back?

client/app/components/AccountConnectionForm.jsx
```diff
@@ -19,25 +19,32 @@ const formConfig = ({ t, fields, accountName }) => {
   }
 }
 
-const AccountConfigForm = ({ t, fields, dirty, submit, submitting }) => (
-  <div class='account-form'>
+const AccountConnectionForm = ({ t, fields, dirty, error, submit, submitting }) => (
+  <div class={'account-form' + (error === 'bad credentials' ? ' error' : '')}>
```
CPatchane: Is it possible to just check the error variable presence here instead of the string content in order to be more extensible? At least just for the form class definition here. We should never have something in the error variable if all goes fine.

CPatchane:
Great @goldoraf ! 👍

PR #630, opened 2016-12-16
Dropdown field type

CPatchane:
This PR:

* [x] Add documentation about a dropdown field type
* [x] Add dropdown field type feature
* [x] Add fallback for frequency field default dropdown values


goldoraf:
Looking good, thanks @CPatchane!

PR #636, opened 2016-12-26
Adapt MAIF connector from master

gregorylegarec:
MAIF connector is an OAuth connector which is not working like `Facebook Events`, `Google Contacts`, or event `EDF`

As `EDF` works on its own, `MAIF` introduces a new way of connecting to an OAuth provider, by specifying a `connectUrl` property.

This PR :
* Handle the `connectUrl` property and render it in an `AccountDialogForm`.
* If the `connectUrl` property contains the sting `redirect_uri`, it adds a redirection URL at the end of `connectUrl`. This is pretty dirty but it's the actual behaviour on master, and I think this is no time to improve that.
* Hide the default "Connect" button if the connector contains a `connectUrl` property.
* Get back the route `/konnectors/:konnectorId/:accountId/redirect` to be able to handle the value return by the `MAIF` OAuth authentication.

A pretty hell to test.


gregorylegarec:
Ping @jacquarg for information :)

PR #637, opened 2016-12-28
Advanced connector configuration

goldoraf:
This PR adds advanced configuration for already connected accounts, plus the ability to handle multiple accounts.


client/app/containers/ConnectorManagement.jsx
```diff
@@ -0,0 +1,216 @@
...
+        type: 'text',
+        advanced: true
+      }
+    }
+    if (fields.frequency && !fields.frequency.default) {
+      fields.frequency.default = 'week'
+    }
+    if (fields.frequency &&
+      (!fields.frequency.options || !fields.frequency.options.length)) {
+      fields.frequency.options = ['hour', 'day', 'week', 'month']
```
CPatchane: Is this default options used in the frequency dropdown field? I didn't see it on the [related component](https://github.com/cozy-labs/konnectors/pull/637/files#diff-aa3d3e066538ff6abfbd2642ff1494b7R35).

client/app/components/AccountConfigForm.jsx
```diff
@@ -0,0 +1,56 @@
...
+    {fields.calendar &&
+      <div>
+        <h3>{t('my_accounts calendar')}</h3>
+        <p>{t('my_accounts calendar desc')}</p>
+        <Field label={false} {...fields.calendar} />
+      </div>
+    }
+    <h3>{t('my_accounts frequency')}</h3>
+    <p>{t('my_accounts frequency desc')}</p>
+    <Field label={false} {...fields.frequency}>
```
CPatchane: Did it make sense to use the dropdown field component here, instead of the simple field? I think it's more appropriate to the frequency fields, what do you think?

client/app/components/Field.jsx
```diff
@@ -93,17 +93,21 @@ export const PasswordField = translate()(
...
+  const { value, options, onChange, onInput } = props
 
   return (
     <FieldWrapper {...props}>
-      <select className='account-field-dropdown'>
-        {options.map(optionValue => {
-          return <option
+      <select
+        className='account-field-dropdown'
+        value={value}
```
goldoraf: This is a React thing. See https://facebook.github.io/react/docs/forms.html#the-select-tag

client/app/containers/ConnectorManagement.jsx
goldoraf: No, it's not used because we need to display the translations, and DropdownField doesn't translate.

client/app/components/AccountConfigForm.jsx
goldoraf: Again, DropdownField doesn't translate its options, and I didn't want to write a I18nDropdownField or something like that. Field knows how to handle children inputs, so be it.

client/app/containers/ConnectorManagement.jsx
CPatchane: So we can remove this part to remove confusion about default values of the frequency field here, what do you think?

goldoraf: Sure ;)

CPatchane:
Great job @goldoraf 👍 ! Thanks.

PR #642, opened 2017-01-02
Add data types definition in connectors

CPatchane:
* [x] Documentation
* [x] Connectors update

### DIFF
--- a/docs/client-side-connector-configuration.md
+++ b/docs/client-side-connector-configuration.md
@@ -167,3 +167,47 @@ Some fields expect to have default values. If it's not the case, fallbacks will 
 * frequency:
     * `'weekly'` as default values
     * `['hourly', 'daily', 'weekly', 'monthly']` as default options
+
+## Data types
+
+A connector must specify which kinds of data it will be able to retrieve from its related service. These types are used to inform the user (information displayed on the connector modal) about his retrieved data when using a connector. Here is an example of data types declaration:
+
+```javascript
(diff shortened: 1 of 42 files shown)


ZeHiro:
Isn't dataType redeundant with the docType defined by the model attribute? Or is this a way to give more sense to the model ? Shouldn't there be a default value for each model? Couldn't there a extra test for that ?

CPatchane:
@ZeHiro It could be seen a little redundant but it's not the same thing. We use dataType here to specify which kind of data will be retrieved by the connector, in a more global, descriptive and semantic way than the docType. 

Here are some differences for example:
* Ameli uses `Bill` model but it's `refund` data type
* Bouygues connectors uses `PhoneBill` and `InternetBill` models for the same global `bill` data type.

I think it's more useful to keep separate the global kind of data and the data models used by the connector in order to be more flexible for data models usage while keeping the data type list exhaustive. 

About default values, I personally don't see a type that could be used as 'default value' :/. Also, if we don't use default values here, that will make the connector's developer define data types 'by himself'.

And about test, I agree with you, we could use a test here to be sure a connector define at least one data type because a connector without declared data types doesn't make any sense (or is a bad connector). I will add it so. 👍 

Hope that respond your questions and thanks for your feedbacks 👍

ZeHiro:
@CPatchane  Yes answers all my questions. Thank you for the tests :)

PR #644, opened 2017-01-02
Fix the EN translation for Ameli Konnector

Darcidride:
The EN translation is false because the "Bank identifier" field is
required, and not optional as described in the EN translation.


ZeHiro:
Thanks for opening the PR, I have created another one with more precise label. Also for next time, can you PR without the built app? Thanks again.

Darcidride:
@ZeHiro Ok thank you for your more detailed PR
For the build app, ok no problem I won't send it the next time

PR #648, opened 2017-01-03
ADD DataTypes views in connectors

GoOz:
- [x] DataTypes from connectors are displayed
- [x] SVGs colored according to the hex value in connectors' config
- [x] DataTypes descriptions in locale

⚠ Must be merged **AFTER** #642 as it depends on it. ⚠


client/app/assets/sprites/icon-activity.svg
```diff
@@ -0,0 +1,16 @@
...
+  <g fill="none" fill-rule="evenodd" transform="translate(-96 -96)">
+    <g stroke-width="2" transform="translate(97 97)" stroke-linecap="round" stroke-linejoin="round">
+      <g stroke="#95999D" transform="translate(0 16)">
+        <path d="M0.0254,30 L46.0254,30"/>
+        <polygon points="8.026 20 2.026 20 2.026 30 8.026 30"/>
+        <polygon points="20.026 10 14.026 10 14.026 30 20.026 30"/>
+        <polygon points="32.026 14 26.026 14 26.026 30 32.026 30"/>
+        <polygon points="44.026 0 38.026 0 38.026 30 44.026 30"/>
+      </g>
+      <g stroke="currentColor" transform="translate(3)">
```
CPatchane: Interesting, I didn't know that we could use the CSS `color` property here using `currentColor`. Good to know 👍

client/app/locales/en.json
```diff
@@ -185,5 +185,28 @@
   "my_accounts disconnect button": "Disconnect this account",
   "my_accounts delete button": "Delete this account",
   "my_accounts account delete success": "Account removed succesfully",
-  "my_accounts account delete error": "Apologies, our server had an hiccup, do you mind starting again?"
+  "my_accounts account delete error": "Apologies, our server had an hiccup, do you mind starting again?",
+
+  "dataType title": "Your Cozy retrieve the following data:",
```
GoOz: True. I copied and pasted the trello card without really reading it. 👅

client/app/assets/sprites/icon-activity.svg
GoOz: Yeah, it didn't come to my mind at first, but now i think it's a far better approach than my previous one. 😄

gregorylegarec:
Looks good to me. 👍 

However, I am asking myself about our previous choices for colors. We store two values : `css` and `hex`. In most of times, those two values are the same. First remark : why do we store a sharp character `#` in the hex value ? This pure detail but my second point is more important : in the css value we store hexadecimal color or background property, making this value unusable as it, without testing what it contains. I think we should improve this to something like :
```json
color: {
  hex: {
    '#9E0017'
  },
  css: {
    background: 'linear-gradient(90deg, #EF0001 0%, #9E0017 100%)'
  }
}
```
So the css value could be computer easily with a simple reduce method.

What do you think @GoOz, @CPatchane, @m4dz, @goldoraf ?

GoOz:
I agree, the # character seems redundant.
As for what is in css value… i don't really know how it is used but if you can have different kind of values, it may be more secure to test it, i guess.

gregorylegarec:
After discusssing with @CPatchane :
* an `hex` value ready-to-use is simpler and easier
* At this time the `css` value is only used for `backgroud` property, so it's not useful to improve it. Maybe the `css` property should have been named something like `css-background` but again, at this this time it is not worth the change.

PR #649, opened 2017-01-04
Quickfix on dataTypes

GoOz:
Fixed a typo, a mix-up in colored path in a svg and wrong filenames & better .gitignore for locales files


CPatchane:
👍

PR #651, opened 2017-01-04
[Fix] connector's description position + locales

GoOz:
Minor fixes


client/app/components/AccountConnection.jsx
```diff
@@ -19,8 +21,7 @@ const AccountConnection = ({ t, connector, connectUrl, fields, dirty, error, sub
             />
           )}
         </ul>
-        <p>{connector.name + t('dataType disclaimer')}</p>
-        <p>{t(description)}</p>
+        <p>{connector.name + ' ' + t('dataType disclaimer')}</p>
```
m4dz: Shouldn't we use a template string here?

```jsx
<p>{`${connector.name} ${t('dataType disclaimer')}`}</p>
```

GoOz: Oh, I didn't know that one. It's far more elegant this way. Thanks.

PR #652, opened 2017-01-04
Add connector function to remove invalid declared data type

CPatchane:
This PR add a function to check connector dataType to remove unlisted/unauthorized types.

Should we add a test for that? I am afraid that will make unit tests much longer to be run. The thing is that if a connector only declares one type, but an invalid one, all tests will be green but no types will be displayed on this connector.

What's your opinion about that?

### DIFF
--- a/client/app/containers/ConnectorManagement.jsx
+++ b/client/app/containers/ConnectorManagement.jsx
@@ -6,6 +6,13 @@ import AccountConnection from '../components/AccountConnection'
 import AccountManagement from '../components/AccountManagement'
 import Notifier from '../components/Notifier'
 
+let AUTHORIZED_DATATYPE = [
+  'activity', 'heartbeat', 'calendar', 'commit',
+  'consumption', 'contact', 'contract', 'travelDate', 'event', 'bill',
+  'stepsNumber', 'podcast', 'weight', 'bloodPressure', 'appointment',
+  'refund', 'sleepTime', 'courseMaterial', 'temperature', 'tweet'
+]
+
 const prepareConnectURL = (connector) => {
   let connectUrl = connector.connectUrl
   if (!connectUrl) {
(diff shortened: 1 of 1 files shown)


client/app/containers/ConnectorManagement.jsx
```diff
@@ -187,6 +194,20 @@ export default class ConnectorManagement extends Component {
     return defaults
   }
 
+  checkProperties (connector) {
+    // remove invalid dataType declaration
+    // (valid = present in AUTHORIZED_DATATYPE)
+    let checkedTypes = []
+    for (let dType of connector.dataType) {
+      if (AUTHORIZED_DATATYPE.find(function (t) {
```
gregorylegarec: I was looking at this PR, so I share my thought : `Array.find()` is a little bit expensive here for just testing an equality. We may use `Array.indexOf` instead. We may also use a function to avoid the comment `// (valid = present in AUTHORIZED_DATATYPE)`, like this one :
```javascript
const isValidType = (type) => AUTHORIZED_DATATYPE.indexOf(type) !== -1
```
We may also go a little further, because what we want to do in `checkProperties` is just __filter__ the valid dataTypes, and because we do not only _check_ but also update the connector, we may call the function _sanitize_ :
```javascript
sanitize (connector) {
  connector.dataType = connector.dataType.filter(isValidType)
  return connector
} 
```
Finally, to improve function purity and connector immutability, I think we may return another connector (I assume that it is ok with the way preact works:
```javascript
sanitize (connector){
  return Object.assign(
    {},
    connector,
    {
      dataType: connector.dataType.filter(isValidType)
    }
  }
```

What do you think ?

CPatchane: Great! I was wondering how to improve this piece of code, here it is 👍.

I also agree about the `sanitize` function, I just not wanted to name a function specific to data type here, we could use it to check other properties later, so it's the right name for here.

 Thanks @gregorylegarec for your feedbacks 🎉. I will do these changes.

goldoraf:
:+1:

ZeHiro:
@cpatchane i think tests over the existance of datatypes for each datatype + existance of translation keys would really help developpers and mainteners of konnectors. I think each time the app exports a constraint to the konnector developper, a test should be added. Of course it will take extra time, but the tests are run 'offline'.

PR #653, opened 2017-01-05
Update Orange_VOD for API changes

jacquarg:
Sorry, the first version of Orange_VOD isn't available yet, but the first API changes are already there.


ZeHiro:
@jacquarg Thanks

PR #658, opened 2017-01-07
Fix captain train konnector

ZeHiro:


nicofrand:
What does it solve ? Mine does not seem to have any issue…

ZeHiro:
Multi trips bills

server/konnectors/captain_train.js
```diff
@@ -228,7 +228,7 @@ function fetchBills(requiredFields, entries, data, next) {
     let linkedPNR = [data.pnrs.find(pnr => pnr.id === proof.pnr_id)];
     try {
       linkedPNR = data.pnrs.filter(
-        pnr => pnr.proof_ids.indexOf(proof.id) !== -1
+        pnr => typeof pnr.proof_ids !== 'undefined' && pnr.proof_ids.indexOf(proof.id) !== -1
```
nicofrand: Do you expect an array ? Maybe `instanceof Array` would be easier to read than !== 'undefined' ?

server/konnectors/captain_train.js
```diff
@@ -97,7 +97,7 @@ function login(requiredFields, entries, data, next) {
         logger.error(err);
         return next(err);
```
nicofrand: Maybe you could return next('request error') here instead ?

server/konnectors/captain_train.js
ZeHiro: Yes.

server/konnectors/captain_train.js
ZeHiro: Of course

nicofrand:
LGTM !

ZeHiro:
@nicofrand I let you merge then

Issue #655, opened 2017-01-06
Wrong year in filenames for Ameli Konnector


nicofrand:
A user reports that a bill from the 22th december 2016 was imported as 20171222 in its filename.
This is due to https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/ameli.coffee#L147 that automatically selects the current year (since it is not provided by the website itself) without checking that this date is not superior to the current date.

ZeHiro:
Closed by #664

PR #664, opened 2017-01-09
Fix year bug in Ameli konnector

Gara64:
The Ameli konnector was using the current year for its imports. So 2017 caused some troubles :)
References #655

### DIFF
--- a/server/konnectors/ameli.coffee
+++ b/server/konnectors/ameli.coffee
@@ -134,6 +134,10 @@ parsePage = (requiredFields, healthBills, data, next) ->
             # Each bloc represents a month that includes 0 to n reimbursement
             $('.blocParMois').each ->
 
+                # It would be too easy to get the full date at the same place
+                year = $($(this).find('.rowdate .mois').get(0)).text()
+                year = year.split(' ')[1]
+
                 $('[id^=lignePaiement' + i++ + ']').each ->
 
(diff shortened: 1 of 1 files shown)


server/konnectors/ameli.coffee
```diff
@@ -134,6 +134,10 @@ parsePage = (requiredFields, healthBills, data, next) ->
             # Each bloc represents a month that includes 0 to n reimbursement
             $('.blocParMois').each ->
 
+                # It would be too easy to get the full date at the same place
```
ZeHiro: ❤️ the comment

server/konnectors/ameli.coffee
```diff
@@ -142,7 +146,7 @@ parsePage = (requiredFields, healthBills, data, next) ->
 
                     month = $($(this).find('.col-date .mois').get(0)).text()
```
ZeHiro: don't you need to do month.split(' ')[0] to get the month ? As for year.

Gara64: No, because there are 2 date locations : one with month and year concatenated (thus, the split is needed) and another one with month and day on their own field (.mois vs .jour). Hence my comment above :)

ZeHiro: Nikel. I merge after fixing master branch.

PR #679, opened 2017-01-17
[fix] Fix test if this is a multiline account

### LINKED ISSUE #672: [free mobile] Wrong management of multi lines account
A user reports it has several bills linked to one bank operation.

ZeHiro:
Should fix #672. This is not tested (I do not have multiline for Free mobile). So this is a blind fix.


nicofrand:
Thanks !

PR #690, opened 2017-02-02
[fix] missing plural locale

GoOz:


client/app/locales/en.json
```diff
@@ -6,7 +6,7 @@
   "notification homes": "%{smart_count} home description imported |||| %{smart_count} home descriptions imported",
   "notification consumptionstatements": "%{smart_count} new consumption statement imported |||| %{smart_count} new consumption statements imported",
   "notification contracts": "%{smart_count} new contract imported |||| %{smart_count} new contracts imported",
-  "notification clients": "%{smart_count} new client's informations imported",
+  "notification clients": "%{smart_count} new client's information imported |||| %{smart_count} new clients' information imported",
```
y-lohse: I believe the second apostrophe is in the wrong place?

GoOz: Nope, it means "clients's information" and you don't do that in english. In that case the possessive "s" disappears.

enguerran: @y-lohse in fact, when you have a plural form, you had a single quote (to mark possession) after the -s that marks the plural form. As explained [here](https://github.com/cozy/cozy-files-v3/pull/56/files#r98928250)

PR #691, opened 2017-02-03
[Fix] Review RWD for connectors item so it doesn't spread too much

GoOz:


gregorylegarec:
Thanks @GoOz, two remarks :

First, in responsive mode, the heading's horizontal margin is too important and should be decreased to `1em`:
```stylus
[role=application] h1 {
    font-size: 2em;
    margin: 1.5em 1em 1em;
}
``` 

Second, I think we should be able to improve the UI by having connectors cards growing as the the width of the viewport change. I know also that it was what @kelukelu was wanting when he designed the page. So we can add this behavior by changing the flex property of `.item-wrapper` to:
```stylus
.item-wrapper
    flex             0 1 22em
    flex-grow   1
```

It works pretty well except for the last line. Good news and bad news here : the good one is that I have a solution, the bad one is that you will pretty sure hate it: to force cards on the last row to stick left and resize theirselves according to cards on other rows, we can add "ghost" element at the end of the list (five should be enough to prevent large screens with six cards per row). This could be done in `ConnectorList.jsx`:
```jsx
const ConnectorList = ({ t, connectors, showConnectedBadge = true }) => (
  <div class='connector-list'>
    {connectors.map(c =>
      <ConnectorItem
        title={c.name}
        subtitle={t(c.category + ' category')}
        connected={showConnectedBadge && c.accounts.length !== 0}
        iconName={c.slug}
        slug={c.slug}
        enableDefaultIcon
        backgroundCSS={c.color.css}
      />
    )}
    <div class='item-wrapper growth-hacker' />
    <div class='item-wrapper growth-hacker' />
    <div class='item-wrapper growth-hacker' />
    <div class='item-wrapper growth-hacker' />
    <div class='item-wrapper growth-hacker' />
  </div>
)
```
and adding the following CSS rule in index.styl:
```stylus
.item-wrapper
    flex             0 1 22em
    flex-grow:       1
// [...]
    &.growth-hacker
        height         0
        padding        0
        margin-top     0
        margin-bottom 0
```
By having their height, margin-top and margin-bottom set to 0, those additionnal elements will not  stack vertically and will not add unexpected bottom margin. 

It works well in Firefox and Chrome, I don't know if it is a known hack but it can do what we are expecting for. Give it a try !

GoOz:
@gregorylegarec Actually, elements had fluid width before but it wasn't what PO wanted so that's what this fix is all about.
And about your idea with 'ghost elements', that's already what I'm doing with ::before and ::after pseudo-elements, but in a much cleaner way 👅

client/app/styles/index.styl
```diff
@@ -99,28 +104,31 @@ aside
...
 
 @media (max-width: (1023/basefont)rem)
     .connector-list, .use-cases-list
         margin 0 1em 1em
 
+$item-margin = .7em
+
 .item-wrapper
-    flex             0 1 22em
+    flex             1 1 'calc(100% - %s * 2)' % $item-margin
```
gregorylegarec: Great ! :1st_place_medal:

client/app/styles/index.styl
```diff
@@ -99,28 +104,31 @@ aside
...
+    justify-content  flex-start
     margin           0 3em 2em
 
-    &::before
-    &::after
-        content    ''
-        flex       0 1 22em
-        order      2
-        margin     .7em
+    // &::before
```
gregorylegarec: Remaining comment

client/app/styles/index.styl
```diff
@@ -158,6 +166,30 @@ aside
         bottom   1em
         right    1em
 
+@media (min-width: 40em)
```
gregorylegarec: Much cleaner 🎉

gregorylegarec:
Thanks GoOz ! 🏆

PR #712, opened 2017-02-20
Fix encryption of passwords for konnectors with object format for fields

nono:
cc: @poupotte

### DIFF
--- a/server/models/konnector.coffee
+++ b/server/models/konnector.coffee
@@ -91,9 +91,10 @@ Konnector::removeEncryptedFields = (fields) ->
 
     for account in @accounts
         passwords = {}
-        for name, type of fields when type is "password"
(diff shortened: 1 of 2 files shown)


nono:
@doubleface the contributing guide says:

> you do not need to build app to submit a PR

Am I missing something?

doubleface:
Oh sorry, I suppose I am missing something. So the build is done when the pullrequest is accepted ?

doubleface:
Since this pullrequest is for the core and not a konnector, I supposed it would be different. I is just easier to test it directly in a cozy when the build part comes with the pullrequest.

nono:
I think that @aenario builds just before publishing on npm

poupotte:
It's ok for me :+1:

PR #719, opened 2017-02-28
Create konnector_declaration_v3.md

doubleface:
Here is a proposition of declaration of konnectors to the new cozy stack (Gozy, v3)

### DIFF
--- a/docs/konnector_declaration_v3.md
+++ b/docs/konnector_declaration_v3.md
@@ -49,12 +49,10 @@ remarks are welcome.
       "bills": {
         "description": "Required to save the bills data",
         "type": "io.cozy.bills",
-        "verbs": ["GET", "POST"]
       },
       "files": {
         "description": "Required to save the bills pdf",
         "type": "io.cozy.files",
-        "verbs": ["GET", "POST"]
       },
     },
     "developer": {
--- a/docs/konnector_declaration_v3.md
+++ b/docs/konnector_declaration_v3.md
@@ -28,6 +28,7 @@ remarks are welcome.
 {
   "version": "1.0.0",
     "name": "Trainline",
+    "type": "node",
     "icon": "./trainline.svg"
     "slug": "trainline",
     "description": "Get all the bill from trainline",
@@ -61,6 +62,9 @@ remarks are welcome.
     }
 }
 ```
+
+The "node" type is the default type of konnector (and the only one at the moment). It may be
+possible in the future to propose weboob konnectors for example.
 
  - Since the konnectors will be needed to be in their own repository, we will propose to the maintainers to create their own repository with
  their konnector or else it will be created in cozy-labs, I suppose.
--- a/docs/konnector_declaration_v3.md
+++ b/docs/konnector_declaration_v3.md
@@ -17,7 +17,7 @@ remarks are welcome.
  - The konnectors will be installed using npm. Any source usable by npm is possible : github, gist, gitlab, bitbucket (git)
    or a tar.gz somewhere on the internet (but not file addresse). More information here : https://docs.npmjs.com/cli/install
  - A konnector package will need the following items :
-    - An index.js which will be run by the stack in a securised container
+    - An index.js (or any file referenced by "main" section of package.json) which will be run by the stack in a securised container
     - The dependencies need to be bundled an built with it (no dependencies are installed by the stack)
     - a cozy-client-js served and updated by the stack will be accessible to the index.js file.
     - shared dependencies (the server/lib and server/models directories) will be copied by the stack in the konnector repository
--- a/docs/konnector_declaration_v3.md
+++ b/docs/konnector_declaration_v3.md
@@ -1,49 +1,57 @@
 # Konnector declaration proposition in V3 (Gozy)
 
 With the cozy architecture change, the konnectors will also need some adaptation to work in the V3. The goal here is to show
-to the community a proposition of declaration of a konnector in the new stack. This proposition is still to be discussed and all 
+to the community a proposition of declaration of a konnector in the new stack. This proposition is still to be discussed and all
 remarks are welcome.
 
 ## Constraints
 
-- For legal reasons, we need to separate the konnectors so that the user needs to decide himself to install a konnector 
+- For legal reasons, we need to separate the konnectors so that the user needs to decide himself to install a konnector
   and not the whole package (like the applications).
-- We want to propose a curated list of konnectors, but we also want to let the user install konnectors from other 
+- We want to propose a curated list of konnectors, but we also want to let the user install konnectors from other
   sources (github, npm, etc)
 - The konnectors will be executed in a container with the last LTS nodejs version (6 at the moment) with cozy-client-js giving access to the cozy stack and with access to internet (of course!)
 
 ## Proposition
 
- - The konnectors will be installed using npm. Any source usable by npm is possible : github, gist, gitlab, bitbucket (git) 
+ - The konnectors will be installed using npm. Any source usable by npm is possible : github, gist, gitlab, bitbucket (git)
    or a tar.gz somewhere on the internet (but not file addresse). More information here : https://docs.npmjs.com/cli/install
  - A konnector package will need the following items :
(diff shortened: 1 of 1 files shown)


nicofrand:
What are the verbs in the permissions ? Why should a konnector developer bother with those ?

ZeHiro:
I agree with @nicofrand : it's these verbs are not really meaningful. read/write is clearer.

nicofrand:
Do you have an example of a konnector that would only need a read access ? Even read/write does not make sense to me.

ZeHiro:
In fact it might be read data from provider or write data to the provider

doubleface:
@ZeHiro about the icons, I supposed it would be in the client application but it may be better to put it and declare it in the konnectors, now that it has its own repository. We "just" have to find a good protocol to serve it to the client. I will look at it.

ZeHiro:
@doubleface if each konnector has its own repo/package, everything should be bundled in the connector. 
@Phyks in its project to integrate weboob to konnectors used based64 images passed throw json API

doubleface:
@ZeHiro @nicofrand About the permission. These are permissions of access to given docTypes. I just used exactly the same permissions as the ones for the applications. You can have more information [in the official documentation](https://cozy.github.io/cozy-stack/permissions.html).

And no, I did not find any existing konnector which would only need read access by it still may be possible. I am working in my spare time on a RSS feeds konnector which would need read access to a list of feeds and read/write access to the list of items. The list of feeds would be updated by a specific application.

But if this type of permissions make contribution more complicated, it may be better to propose some shortcuts like : 

```
"permissions" : ["io.cozy.bills", "io.cozy.files"]
```
Which give all permissions on theses docTypes.
Does it make more sense?

nicofrand:
Yup, the shortcuts would make more sense to me.

ZeHiro:
I would say, as long as the permissions are really clear, I don't really care :)

docs/konnector_declaration_v3.md
```diff
@@ -0,0 +1,63 @@
...
+to the community a proposition of declaration of a konnector in the new stack. This proposition is still to be discussed and all 
+remarks are welcome.
+
+## Constraints
+
+- For legal reasons, we need to separate the konnectors so that the user needs to decide himself to install a konnector 
+  and not the whole package (like the applications).
+- We want to propose a curated list of konnectors, but we also want to let the user install konnectors from other 
+  sources (github, npm, etc)
+- The konnectors will be executed in a container with the last LTS nodejs version (6 at the moment) with cozy-client-js giving access to the cozy stack and with access to internet (of course!)
```
nono: Node 8 will be released in april. I think we will start with it

docs/konnector_declaration_v3.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Proposition
+
+ - The konnectors will be installed using npm. Any source usable by npm is possible : github, gist, gitlab, bitbucket (git)
```
nono: npm or yarn? I have a preference for yarn, but it may be more difficult to use (no global install for example)

docs/konnector_declaration_v3.md
```diff
@@ -0,0 +1,63 @@
...
+- We want to propose a curated list of konnectors, but we also want to let the user install konnectors from other 
+  sources (github, npm, etc)
+- The konnectors will be executed in a container with the last LTS nodejs version (6 at the moment) with cozy-client-js giving access to the cozy stack and with access to internet (of course!)
+
+## Proposition
+
+ - The konnectors will be installed using npm. Any source usable by npm is possible : github, gist, gitlab, bitbucket (git) 
+   or a tar.gz somewhere on the internet (but not file addresse). More information here : https://docs.npmjs.com/cli/install
+ - A konnector package will need the following items :
+    - An index.js which will be run by the stack in a securised container
```
nono: We can use the `main` field from `package.json` and fallback to `index.js` if missing

nono: Cf https://docs.npmjs.com/files/package.json#main

docs/konnector_declaration_v3.md
```diff
@@ -0,0 +1,63 @@
...
+        "bills": {
+          "description": "Required to save the bills data",
+          "type": "io.cozy.bills",
+          "verbs": ["GET", "POST"]
+        },
+        "files": {
+        "description": "Required to save the bills pdf",
+        "type": "io.cozy.files",
+        "verbs": ["GET", "POST"]
+      },
```
nono: Can you fix indentation here?

docs/konnector_declaration_v3.md
```diff
@@ -0,0 +1,63 @@
...
+```
+  {
+    "version": "1.0.0",
+    "name": "Trainline",
+    "slug": "trainline"
+    "description": "Get all the bill from trainline",
+    "source": "https://github.com/cozy/trainline@build",
+    "locales": {
+        "fr": {
+            description: "Récupère toutes vos factures trainline"
```
nono: Can you add the translation of the description of permissions?

nono:
`verbs` is optional. But we want to have in the future some konnectors that can synchronize things from the cozy to external services and a read-only permission (`GET`) makes sense there.

We also think it's nice to have a description for each permission (that can be localized) to explain to the user why the permission is needed.

nicofrand:
If a konnector is now able to write things toward the provider, I'd prefer that the user allows it instead of declaring it in the konnector.

nono:
> If a konnector is now able to write things toward the provider, I'd prefer that the user allows it instead of declaring it in the konnector.

It's the goal. The user must accept the permissions to use the konnectors.

docs/konnector_declaration_v3.md
doubleface: I am not a regular yarn user. yarn.lock is quite fine and we won't need global install from what I know, but ```yarn add``` does not allow install from gitlab or bitbucket (but url of tar.gz files allowed).
Maybe yarn will be better since it is also faster.

docs/konnector_declaration_v3.md
doubleface: OK, will update it

ZeHiro:
@nono how this will be implemented konnector-side ? If the konnector "wants" to push information to a provider, it will just do it. Will the data be pushed by the plateform and not the konnector ?

docs/konnector_declaration_v3.md
doubleface: done

docs/konnector_declaration_v3.md
doubleface: done

doubleface:
@nicofrand @ZeHiro after @nono remark about permissions, I don't think the shortcut I proposed is OK. Since the verbs are optional, I think this is easy enough and we still need to define the text which will be displayed to the user to accept the permissions. So I will remove the "verbs" attribute in my proposition.

doubleface:
@nicofrand @ZeHiro I updated the document with your remarks and also removed npm and use git only instead. Npm could not do version management with git repositories and it is easier to publish on github than on npm (for the majority of users I think).

If you are OK, I will merge this document and publish a post on the forum to introduce it and have a wider range of feedback. Anyway, this document is not definitive at all at the moment.

docs/konnector_declaration_v3.md
```diff
@@ -28,6 +28,7 @@ remarks are welcome.
 {
   "version": "1.0.0",
     "name": "Trainline",
+    "type": "node",
```
nono: Can you also add the `main` in this example?

doubleface: Actually, I supposed the ```main``` section would be in the package.json but that way, we would not have to read the package.json at all and we would run the konnector with a simple node command not using npm. Ok, I put the main section in the manifest.

nono: Sorry, my bad, you are right, it should be in package.json

doubleface: It is not so bad if we don't want to have to read in the package.json to run the konnector. Or we can run the konnector with npm start and let the developper of the konnector handle that part.

PR #718, opened 2017-02-24
Add update route on konnector

jacquarg:
This is a preliminar work to integrate some modification on official Orange Mobile konnector, in the scope of the MesInfos pilote. ping @poupotte 

* Add update route on konnector, 
* in Oauth connectors, save account data before redirect to authent
* automatically import on redirect
* redirect to the connected page.
* update Facebook konnector
* fir #420 in Facebook konnector.

* Could be done : there isn't "everything fine" message with OAuth connector, but just a redirection to connected page: is there a quit fix to improve this ?

PR #720, opened 2017-03-01
Update Orange _mobile and Orange_vod messages

jacquarg:
Update Orange_mobile and Orange_vod messages, with the ones from jurist of orange.
Add opt-in button on Orange_mobile, to activate a Orange geolocation service : it relies on #718 to do that.

* I had to propose a specific layout for checkboxes  (ping @gregorylegarec : it was my first react experience, some proofreading should be useful ! )

* Translation are still messy, but I have some keys to add in transifex.


client/app/components/Field.jsx
```diff
@@ -115,3 +115,47 @@ export const DropdownField = translate()((props) => {
...
+        type='checkbox'
+        className='account-field-input-checkbox'
+        value={value}
+        onChange={onChange}
+        onInput={onInput}
+      />
+      )
+  }
+
+  // ({ required, label, dirty, touched, errors, children }) => {
```
gregorylegarec: ⚠️  Remaining comment

client/package.json
```diff
@@ -5,6 +5,7 @@
     "normalize.css": "5.0.0",
     "preact": "^7.1.0",
     "preact-compat": "^3.9.2",
+    "react-markdown": "^2.4.4",
```
gregorylegarec: I cannot find where it is used, does it need to be added as dependency ?

(ReactMarkdown is imported in `client/app/components/AccountConnection.jsx` but it seems to be never used, we may remove it I think)

server/konnectors/orange_mobile.js
```diff
@@ -31,6 +31,21 @@ const connector = module.exports = baseKonnector.createNew({
   },
```
gregorylegarec: We may add the orange_mobile connector into 'telecom' category, what do you think ?

server/controllers/konnectors.coffee
```diff
@@ -77,6 +50,24 @@ module.exports =
 
             res.status(204).send konnector
 
+    update: (req, res, next) ->
+        # Don't update connector if an import is already running.
+        if req.konnector.isImporting
+            res.send 400, message: 'konnector is importing'
```
gregorylegarec: HTTP status 400 means _Bad Request_. But here the request is correct. Maybe a 409, _Conflict_ sould be more accurate ? What do you think ?

gregorylegarec:
@m4dz Travis tests are failing, because of an untranslated connector field named `frequency`. But none of the concerned connectors (orange_mobile and orange_vod) have this field. Do you have any clue about the reason ? I am totally lost.

client/package.json
jacquarg: ok, I will remove the import, and the dependency.
build wasn't working in master, and npm was complaining about react-markdown, so I tiried the easiest solution (npm install :  ).

client/app/lib/MyAccountsStore.js
```diff
@@ -71,26 +71,39 @@ export default class MyAccountsStore {
...
       .then(response => {
         return response.status === 200
+          ? response.text()
+          : Promise.reject(response)
+      })
+      .then((body) => {
+        let connector = JSON.parse(body)
+        this.updateConnector(connector)
+        return connector
+      })
```
jacquarg: I thought that was already the behaviour of _then_ : if a promise emit a _reject_, it's directly handled by the next _catch_ (or then 2nd argument).

So your proposition is exactly the same thing as the initial code.
Do I miss something ? : )

server/konnectors/orange_mobile.js
jacquarg: Yes definitely.
I put orange_vod in isp category.

server/controllers/konnectors.coffee
jacquarg: I agree, a 409 woold be more accurate, or even a 423 : Locked ; or 429 : Too Many Requests

To be honnest, I just copy past this from the _import_ method (https://github.com/jacquarg/konnectors/blob/61d38892dd43284bca15ecef349fb281d44c07d7/server/controllers/konnectors.coffee#L78-L81 ). We should the both response, but I am not sure that it will not break something in the client.

client/package.json
jacquarg: Ooops, actualy, it's used here : https://github.com/jacquarg/konnectors/blob/61d38892dd43284bca15ecef349fb281d44c07d7/client/app/components/AccountConnection.jsx#L18-L23

jacquarg:
Hello, I made some changes, I hope that I fixed the problem of travais ; I add the "frequency" entry in locales.

m4dz:
@gregorylegarec beacause the field's key is used as its `name` in https://github.com/cozy-labs/konnectors/blob/master/client/app/components/AccountLoginForm.jsx#L28 an so will try to be translated. Need to add a dict entry in `en.json` file for `frequency` (and update it as well on Tx :reminder_ribbon:)

m4dz:
@gregorylegarec is it now ready to merge?

server/controllers/konnectors.coffee
gregorylegarec: You're right. Let's keep it as it for the moment and let's just keep in mind that we should think about those "already importing" status. I like your propositions 👍

client/app/lib/MyAccountsStore.js
gregorylegarec: I verified and you're absolutely right. It seems that it's me who missed something :(

I got some code to clean !

gregorylegarec:
@m4dz Everything's ok, thanks @jacquarg !

jacquarg:
Thank @gregorylegarec  !
Some translation needs update (and some string needs to be added in transifex, and there is some french in english file at this point) .
Could you upload the en.json to transifex, I could then update values in transifex.

Issue #182, opened 2016-02-03
Konnector for Digiposte


Chocobozzz:
Digiposte is a service proposed by "La Poste" (french post office) which is a "Digital Safe" where your employer could drop your payroll there. You can drop your files (photos etc) too.

Moreover, they partnered with many companies that can send documents (like bills, administrative things etc) inside DigiPoste. For example there are Amazon, Ebay, EDF, impots.gouv.fr, CAF, Paypal, OVH...

It seems they have an API https://api.recfonc.u-post.fr/docs/.

clochix:
Thanks @Chocobozzz,
If you develop in Node.js (or would like to learn), we would be glad to help you create this Konnector. Don't hesitate to ask!

Chocobozzz:
Hi @clochix :)

No problem, when I'll have time I'll try to make it. But it can be long, my hollidays will be in a long time. I'll notify it if I begin the work ;)

clochix:
Awesome!

PR #724, opened 2017-03-08
New digiposte konnector

doubleface:
Here is a new digiposte konnector, which only gets bills at the moment but it's not too bad I suppose.

Fixes also #182


build/server/konnectors/digiposte.js
```diff
@@ -0,0 +1,230 @@
...
+    connector.logger.info('Number of fetched documents : ' + nb);
+    if (nb > 0) {
+      var localizationKey = 'notification bills';
+      var options = {
+        smart_count: nb
+      };
+      next(null, localization.t(localizationKey, options));
+    }
+  }).catch(function (err) {
+    next(err);
```
nicofrand: That would be better to log the error and return the actual locale key corresponding to the error. Ex: 'bad credentials'.

doubleface: This should be fixed now

nicofrand: There should be a parsing error here too : https://github.com/doubleface/konnectors/blob/e4228f3f0d3fa1542910a420a4902e554347ed60/server/konnectors/digiposte.js#L57 and here https://github.com/cozy-labs/konnectors/pull/724/files/6dfb55e3565604a6c473b5dedb24ca8b517d958b#diff-dab0f39cee0d1813f8adc9f00c8b9b78R109

doubleface: Wow fast review. Thank you!

This should be fixed now

nicofrand: BTW, we usually don't commit the build directory directly, is there a change inside cozy team ?

build/server/konnectors/digiposte.js
```diff
@@ -108,7 +111,8 @@ function fetchBills(requiredFields, bills, data, next) {
       connector.logger.info('App access token is ' + accessToken);
       return accessToken;
     } else {
-      return Promise.reject('Problem fetching the access token');
+      connector.logger.error('Problem fetching the access token');
+      return Promise.reject('parsing error');
```
nicofrand: There is a dedicated locale key for that : https://github.com/cozy-labs/konnectors/blob/master/client/app/locales/en.json#L12

server/models/file.coffee
```diff
@@ -93,10 +94,17 @@ File.createNew = (fileName, path, url, tags, callback) ->
...
 
     # Save file in a tmp folder while attachBinary supports stream.
-    options =
-        uri: url
-        method: 'GET'
-        jar: true
+    options = {}
+    if requestoptions
+        options = requestoptions
+        if _.isFunction options.form
```
ZeHiro: Why not using 'typeof options.form === ' function' instead ?

ZeHiro: BTW, love the feature :)

build/server/konnectors/digiposte.js
```diff
@@ -0,0 +1,243 @@
+'use strict';
```
ZeHiro: Con you not push a built version of the K ? instead the rules have changed : - )

server/konnectors/digiposte.js
```diff
@@ -0,0 +1,249 @@
...
+  },
+  dataType: ['bill'],
+  models: [Bill],
+  fetchOperations: [fetchBills]
+})
+
+function fetchBills (requiredFields, bills, data, next) {
+  let request = require('request-promise')
+  const j = request.jar()
+  // require('request-debug')(request)
```
ZeHiro: Forgotten comment

build/server/konnectors/digiposte.js
doubleface: Ok, but should I push a built version of file and save_data_and_files ?

doubleface: I let you finish the review and will do a new pullrequest without the built files. Please don't merge it.

server/models/file.coffee
doubleface: Juste because lodash is already here and it feels more explicit

server/konnectors/digiposte.js
doubleface: That was on purpose, actually. It is here to make the debugging easier.

server/models/file.coffee
nicofrand: or just `instanceof Function` :p ?

doubleface: Ok, ok :)

server/konnectors/digiposte.js
doubleface: And this module is cool for developing new konnectors

doubleface:
Made a new pullrequest without the build files : #725

Issue #730, opened 2017-03-13
[podcast] Cannot read property 'Symbol(Symbol.iterator)' of undefined


clochix:
Seen in some error logs:
```
TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/build/server/konnectors/podcast.js:283:34
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/lib/model.js:211:20
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/lib/cozymodel.js:290:18
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/lib/model.js:211:20
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/lib/cozymodel.js:290:18
```

Fix in progress.

Issue #729, opened 2017-03-13
[podcast] Cannot read property 'title' of undefined


clochix:
Seen in some error logs:
```
Mar 13 10:20:36 localhost.localdomain konnectors[27424]: /usr/local/cozy/apps/konnectors/node_modules/konnectors/build/server/konnectors/podcast.js:352
          throw _iteratorError4;
          ^

TypeError: Cannot read property 'title' of undefined
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/build/server/konnectors/podcast.js:337:24
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/lib/model.js:213:18
    at /usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/lib/cozymodel.js:294:18
    at parseBody (/usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/node_modules/request-json-light/main.js:75:10)
    at IncomingMessage.<anonymous> (/usr/local/cozy/apps/konnectors/node_modules/konnectors/node_modules/cozydb/node_modules/request-json-light/main.js:109:14)
```
(fix in progress)

PR #731, opened 2017-03-13
Fix some crashes

clochix:
Res #729 et #730

### DIFF
--- a/server/konnectors/podcast.js
+++ b/server/konnectors/podcast.js
@@ -242,9 +242,11 @@ function createFileIfNotPresent (filename, path, url, callback) {
       callback(err)
     }
 
-    for (const file of files) {
-      if (file.name === filename && file.path === path) {
-        return callback(null, file)
+    if (files) {
+      for (const file of files) {
+        if (file.name === filename && file.path === path) {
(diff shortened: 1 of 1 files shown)


server/konnectors/podcast.js
```diff
@@ -242,9 +242,11 @@ function createFileIfNotPresent (filename, path, url, callback) {
       callback(err)
```
ZeHiro: Can you also `return callback(err)` here ? 
This looks like the origin of the problem is here.

server/konnectors/podcast.js
```diff
@@ -273,8 +275,8 @@ function createTrackIfNotPresent (trackName, fileID, callback) {
 
     for (const track of tracks) {
```
ZeHiro: I think the origine of the problem is here : https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/podcast.js#L271

can you `return callback(err)` instead ?

Same here : https://github.com/cozy-labs/konnectors/blob/master/server/konnectors/podcast.js#L226
 Thanks @clochix

ZeHiro:
Thanks @clochix

PR #733, opened 2017-03-16
Updated webpack/babel config so that older browsers without Object.assign get support

goldoraf:


client/app/index.jsx
```diff
@@ -1,4 +1,5 @@
 import './lib/polyfills'
+import 'babel-polyfill'
```
m4dz: We should probably remove lib/polyfills as the polyfills are auto-loaded via through Babel now?

goldoraf: Oh right, I forgot to check that and to remove these polyfills. Good catch, thanks ;)

goldoraf: @m4dz : done! Are we good now according to you?

PR #734, opened 2017-03-16
Do identification in a popup for OAut-h konnector

jacquarg:
, to avoid troubles with http redirect during Orange_mobile and orange_vod connectors.


goldoraf:
Thanks a lot :+1:

PR #743, opened 2017-05-29
fix: konnector ameli

enguerran:
There is a change in ameli webpage and the connection was broken.
Here is a new check with cheerio parsing.


doubleface (approved):

Gara64:
Ah, as simple as that... Thanks @enguerran!

paultranvan (approved):

